PR #7, opened 2017-06-13
Create client.es.yml

SidVal:
I looked in transifex and I couldn't find anything about client.es.yml
So, I created. :es:


techAPJ:
👍

PR #11, opened 2017-12-18
Update client.en.yml

SidVal:
Before: "There are no queries. Why not "
After: "There are no queries. Why not create one?"


SidVal:
My bad. 
Ignore this.

PR #12, opened 2018-05-10
copyediting

robbyoconnor:


jomaxro:
The current copy appears correct to me.  The complete sentence is `Please select or create one.` so the `c` in create should indeed be lowercase.

robbyoconnor:
I guess the question would be why isn't the whole sentence here?

jomaxro:
The sentence is broken up because only "create one" is a link. If the translation was a single sentence we'd have to make the entire sentence a link.

robbyoconnor:
nevermind then :)

PR #13, opened 2018-08-12
[WIP] FEATURE: List previous queries on Data Explorer homepage

rishabhnambiar:
- [x] List previous queries with new details like last run date and user
- [x] Store new fields like user, time in PluginStore
- [x] Style according to Kris's mockup, show/hide queries when appropriate
- [x] Make rows **clickable** to open Query details
- [x] **Sort** list of queries according to run date (id), currently gets sorted by query name (???)
- [x] Figure out how to not hard code user, maybe `current_user.username`
- [ ] sort by “creator, name, last run, created”, click columns
- [ ] Add user profile picture in the `Created by` column

dataex


plugin.rb
```diff
@@ -567,12 +567,15 @@ def self.denormalized_columns
...
   class DataExplorer::Query
-    attr_accessor :id, :name, :description, :sql
+    attr_accessor :id, :name, :description, :sql, :user, :time
 
     def initialize
       @name = 'Unnamed Query'
       @description = 'Enter a description here'
       @sql = 'SELECT 1'
+      #TODO: Figure out where to assign current user for storage
+      @user = 'Admin'
```
rishabhnambiar: @techAPJ 
I need to send the current user's name through the Serializer. 
Where would be a good place to assign it? `current_user.username` does not seem to work inside initialize()

plugin.rb
```diff
@@ -567,12 +567,15 @@ def self.denormalized_columns
 
   # Reimplement a couple ActiveRecord methods, but use PluginStore for storage instead
   class DataExplorer::Query
-    attr_accessor :id, :name, :description, :sql
+    attr_accessor :id, :name, :description, :sql, :user, :time
```
techAPJ: Instead of `user` and `time` we should use `created_by` and `created_at` for clarity.

plugin.rb
techAPJ: Instead of initializing query with user and time, I think we should pass them on create via controller.

rishabhnambiar: @techAPJ, so I should remove all occurrences of user and time from this file and write a new component which will set time and user? Any relevant example for this?

rishabhnambiar: @techAPJ, so I should remove all occurrences of user and time from this file and write a new component which will set time and user? 
Or do you only want me to remove user and time from initialize()?

techAPJ: Serializer is used to pass data to the client (front-end). In this case (saving user and timestamp in a row) everything can happen on server (back-end) because server already knows the current (logged in) user. What you need to do is that pass the current user and timestamp when the (rails) controller receives create request [here](https://github.com/discourse/discourse-data-explorer/blob/master/plugin.rb#L937-L945).

No need to initialize the empty query with user_id and timestamp. Note that how `id` is allocated on runtime. Also instead of `username` we should save user `id` in `created_by` field.

rishabhnambiar: Putting the definition in create makes so much sense, understood!
I've fixed that, thank you!

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
```diff
@@ -78,6 +82,10 @@ export default Ember.Controller.extend({
       this.set('showCreate', true);
     },
 
+    showRecentQueries() {
```
techAPJ: Where is this action being used?

plugin.rb
```diff
@@ -567,7 +567,7 @@ def self.denormalized_columns
 
   # Reimplement a couple ActiveRecord methods, but use PluginStore for storage instead
   class DataExplorer::Query
-    attr_accessor :id, :name, :description, :sql
+    attr_accessor :id, :name, :description, :sql, :created_by, :created_at, :username
```
techAPJ: I don't think we need to save both `id` and `username`. We should just save `id` and user should be fetched in serializer to pass on to client.

plugin.rb
```diff
@@ -609,7 +609,7 @@ def self.alloc_id
 
     def self.from_hash(h)
       query = DataExplorer::Query.new
-      [:name, :description, :sql].each do |sym|
+      [:name, :description, :sql, :created_by, :created_at, :username].each do |sym|
```
techAPJ: No need to add new fields here since it's not supplied by client.

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
rishabhnambiar: Oh, I left it there accidentally :+1:

plugin.rb
rishabhnambiar: I thought so too, but whenever I remove these fields from `from_hash` and `to_hash`, the values don't show up anymore even after creating a new query. That's why I left them there.

I just tried this again and removing the fields seems to break the table values.

SamSaffron:
Let me know when this is ready to merge! looks fantastic, be sure to add to your list giving https://meta.discourse.org/t/data-explorer-plugin/32566 and update once we are done with enough changes.

plugin.rb
techAPJ: I see that `from_hash` is being used to fetch all queries and pass it on to client, so the fields are indeed needed here. Ignore my previous comment.

plugin.rb
techAPJ: What I meant was do not save `username` at all in the row since we are already saving user id and fetch the user details in serializer and pass it on to client.

So removing `username` from here means removing it from everywhere you added it and fetching user in serializer instead.

Note that this (fetching user in serializer) is already needed in the next step which is to show user avatar on index page.

rishabhnambiar: Oh, I understand now. That makes a lot of sense, I shouldn't have left the extra variable there.

By fetching user details in serializer, do mean something like this and using `{{query.fetch_user.username}}`? 

```
class DataExplorer::QuerySerializer < ActiveModel::Serializer
    attributes :id, :sql, :name, :description, :param_info, :created_by, :created_at

    def param_info
      object.params.map(&:to_hash) rescue nil
    end

    def fetch_user
      object.user = current_user
    end
  end
```
This does not seem to work, is this the correct way to fetch user?

techAPJ: By fetching user details I mean retrieving the user row in serializer.

To get this PR merged the simplest thing we can do right now with minimal changes is:

```ruby
class DataExplorer::QuerySerializer < ActiveModel::Serializer
  attributes :id, :sql, :name, :description, :param_info, :created_by, :created_at, :username

  def param_info
    object.params.map(&:to_hash) rescue nil
  end

  def username
    User.find(created_by).username
  end
end
```

But note that this causes n+1 queries so we need to think of better solution in the next round of changes.

rishabhnambiar:
```ruby
def username
    User.find(created_by).username
end
```
This just didn't work for me, I tried very hard to make it work. I could not get rid of some errors when I used that^

Can you check the current changes? 
I've simply added  `query.username = User.find(query.created_by).username` in def create() and this works as expected for now. Looks okay?

techAPJ: Here is what I meant: https://github.com/discourse/discourse-data-explorer/pull/13/commits/a923fc8af48277bb2f4d92bf28959f3be1d893dc

Can you verify the final changes locally? Also we need to rebase this branch with master.

Once all is done feel free to merge at will. 👍

rishabhnambiar:
> Let me know when this is ready to merge!

@SamSaffron This has been reviewed by Arpit, ready to merge :+1: 
Will start working on the next phase now. (sequential loading, user badges)

PR #17, opened 2018-08-19
FIX: Remove bad default query description and add placeholder instead

rishabhnambiar:
placeholder


rishabhnambiar:
Thanks, @ZogStriP :+1:

PR #19, opened 2018-08-20
FIX: Use predefined SCSS colors instead of hardcoded colors

rishabhnambiar:
exp

cc @jjaffeux, 
I've used the style guide and picked near-identical colors. Looks okay?


rishabhnambiar:
Thanks @ZogStriP :+1:

PR #20, opened 2018-08-27
FEATURE: Sort queries by last_run_at as default

rishabhnambiar:
lastrun

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -567,7 +567,7 @@ SQL
 
   # Reimplement a couple ActiveRecord methods, but use PluginStore for storage instead
   class DataExplorer::Query
-    attr_accessor :id, :name, :description, :sql, :created_by, :created_at
+    attr_accessor :id, :name, :description, :sql, :created_by, :created_at, :last_run_at
 
     def initialize
       @name = 'Unnamed Query'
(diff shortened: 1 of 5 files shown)


plugin.rb
```diff
@@ -938,10 +939,11 @@ def show
 
     def create
       # guardian.ensure_can_create_explorer_query!
-
+      
       query = DataExplorer::Query.from_hash params.require(:query)
-      query.created_at = Time.now.strftime("%b %e, %Y")
+      query.created_at = Time.now
       query.created_by = current_user.id.to_s
+      query.last_run_at = '-'
```
ZogStriP: Feels weird. Why isn't this done in the client-side? This looks like a presentation-only edge case.

techAPJ: I think we should default `last_run_at` to the time query was created instead of keeping it blank.

plugin.rb
```diff
@@ -999,6 +1001,9 @@ def schema
     def run
       check_xhr unless params[:download]
       query = DataExplorer::Query.find(params[:id].to_i)
+      query.last_run_at = Time.now
+      query.save
```
techAPJ: This should be `query.save!`so that we raise an error if any.

plugin.rb
rishabhnambiar: Yeah my bad, defaulting this to `Time.now` makes more sense than this. Thanks!!

plugin.rb
rishabhnambiar: Doing that returns `NoMethodError - undefined method save! for #<DataExplorer::Query:0x000056235ba9abf8>`. 
I believe this is a built-in function, any ideas?

techAPJ: I see, we have implemented the `save` function in `DataExplorer::Query` manually. Ignore.

Just commit the `last_run_at` change we discussed above and feel free to merge the PR.

rishabhnambiar: Will do, thanks! :)

PR #21, opened 2018-08-30
FEATURE: Sort queries by names, last_run_at by clicking table headers

rishabhnambiar:
clicktosort


assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
```diff
@@ -119,6 +119,13 @@ export default Ember.Controller.extend({
       this.save().then(() => this.send('run'));
     },
 
+    sortByProperty(property) {
+      if (this.sortBy == `${property}:desc`){
```
techAPJ: We should use `===` here. Configure [ESLint](https://eslint.org) in your code editor to run on save.

Also for consistent formatting (notice `() {` vs `(){` above), I recommend configuring your code editor to run [Prettier](https://prettier.io/) on save.

assets/javascripts/discourse/templates/admin/plugins-explorer.hbs
```diff
@@ -147,9 +147,18 @@
       {{#if sortedQueries.length}}
         <table class="recent-queries">
           <thead class="heading-container">
-            <th class="col heading name">{{i18n 'explorer.query_name'}}</th>
-            <th class="col heading user">{{i18n 'explorer.query_user'}}</th>
-            <th class="col heading time">{{i18n 'explorer.query_time'}}</th>
+            <th {{action "sortByProperty" "name"}} class="col heading name">
+              <i class="fa fa-sort"></i>
```
techAPJ: We should use `d-icon` here and below. For reference: https://github.com/discourse/discourse/blob/637850d8672855574aea8a43e5e1c1ad40ce0861/app/assets/javascripts/discourse/templates/user.hbs#L48

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
rishabhnambiar: I tried `===` at first, then I realised that `===` can never return true for arrays created in different locations, even if their contents are the same. In this case, I think we can only use `==`. 
More info here: https://stackoverflow.com/a/13757153

Yes, I've set up Prettier and ESLint now :+1:

assets/javascripts/discourse/templates/admin/plugins-explorer.hbs
rishabhnambiar: Done, won't use `<i>` again, thanks! :grin:

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
techAPJ: Why are we comparing an array with a string? This should be:

```
(this.sortBy[0] === `${property}:desc`)
```

rishabhnambiar: Wow, my bad, this is much better :disappointed: 
Fixed :+1:

PR #22, opened 2018-09-05
FEATURE: Enable sideloading of curated queries from stock_queries.json

rishabhnambiar:
- [ ] Decide how to make system queries stand out visually
- [ ] Curate a list of ~20 awesome queries from Meta
- [x] Decide appropriate date/year for `query.created_at`

stock

**How this works:**
If there are no queries in the DB that have been created by the `system` user, a set of curated queries are imported from config/stock_queries.json.

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -919,6 +919,11 @@ SQL
     def index
       # guardian.ensure_can_use_data_explorer!
       queries = DataExplorer::Query.all
+      if !queries.any?{|q| q.created_by == Discourse::SYSTEM_USER_ID.to_s}
+      # if there are no queries created by system user
+        add_default_queries
+      end
+
       render_serialized queries, DataExplorer::QuerySerializer, root: 'queries'
     end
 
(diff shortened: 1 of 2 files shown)


plugin.rb
```diff
@@ -1079,6 +1082,19 @@ def run
         end
       end
     end
+
+    def add_stock_queries
+      json = JSON.parse(File.read(File.expand_path("../config/stock_queries.json", __FILE__)))
+      json["queries"].each do |params|
+        query = DataExplorer::Query.new
+        query.sql = params["sql"]
```
rishabhnambiar: @techAPJ Is there a way to assign these 5 properties in a more concise way? I tried using `assign_attributes` but that didn't seem to work.

plugin.rb
```diff
@@ -919,6 +919,10 @@ def check_enabled
     def index
       # guardian.ensure_can_use_data_explorer!
       queries = DataExplorer::Query.all
+      unless queries.any?{|q| q.created_by == Discourse::SYSTEM_USER_ID.to_s}
+        add_default_queries
```
riking: Take a DistributedMutex here and recheck the condition, otherwise you could end up with multiple copies of the queries created.

rishabhnambiar: Hi @riking! I'll now look into how DistributedMutex works. I didn't quite understand how multiple copies of the queries can be created, can you explain a bit? Thank you!

techAPJ:
@rishabhnambiar instead of checking for stock queries on every index page load I recommend adding a once-off job to seed the queries.

rishabhnambiar:
Thanks for taking the time to review @riking, @techAPJ!
I'm closing this PR for now and exploring a better way to achieve this. :+1:

PR #23, opened 2018-09-07
FIX: Remove useless no_query_hook text and move create & import butto…

rishabhnambiar:
…ns to top right

### Before
beforehook

### After
afterhook

### DIFF
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -61,8 +61,4 @@ en:
       explain_label: "Include query plan?"
       save_params: "Set Defaults"
       reset_params: "Reset"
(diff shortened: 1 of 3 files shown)


config/locales/client.en.yml
```diff
@@ -61,8 +61,4 @@ en:
       explain_label: "Include query plan?"
       save_params: "Set Defaults"
       reset_params: "Reset"
```
rishabhnambiar: @techAPJ These vars are unused so is it safe to delete them from the source file?

techAPJ: Yes, unused vars can be deleted safely.

PR #25, opened 2018-09-17
FEATURE: Show the number of results after running a query (limit 250)

rishabhnambiar:
top250

---

19results

### DIFF
--- a/assets/javascripts/discourse/components/query-result.js.es6
+++ b/assets/javascripts/discourse/components/query-result.js.es6
@@ -34,7 +34,7 @@ const QueryResultComponent = Ember.Component.extend({
 
   @computed('content.result_count')
   resultCount: function(count) {
-    if (count === 250) {
+    if (count === 1000) {
       return I18n.t('explorer.max_result_count', { count });
     } else {
       return I18n.t('explorer.result_count', { count });
--- a/assets/javascripts/discourse/components/query-result.js.es6
+++ b/assets/javascripts/discourse/components/query-result.js.es6
@@ -1,6 +1,7 @@
 import { ajax } from 'discourse/lib/ajax';
 import Badge from 'discourse/models/badge';
 import { getOwner } from 'discourse-common/lib/get-owner';
+import { default as computed } from 'ember-addons/ember-computed-decorators';
 
 function randomIdShort() {
   return 'xxxxxxxx'.replace(/[xy]/g, function() {
@@ -29,8 +30,17 @@ const QueryResultComponent = Ember.Component.extend({
   columns: Em.computed.alias('content.columns'),
   params: Em.computed.alias('content.params'),
   explainText: Em.computed.alias('content.explain'),
-
   hasExplain: Em.computed.notEmpty('content.explain'),
+
+  @computed('content.result_count')
+  resultCount: function(count) {
+    if (count === 250) {
+      return I18n.t('explorer.max_result_count', { count });
+    } else {
+      return I18n.t('explorer.result_count', { count });
+    }
+  },
+
   colCount: function() {
     return this.get('content.columns').length;
   }.property('content.columns.length'),
--- a/config/locales/client.en.yml
+++ b/config/locales/client.en.yml
@@ -52,6 +52,12 @@ en:
       download_csv: "CSV"
       others_dirty: "A query has unsaved changes that will be lost if you navigate away."
       run_time: "Query completed in {{value}} ms."
+      result_count:
+        one: "%{count} result."
+        other: "%{count} results."
+      max_result_count:
+        one: "Showing top %{count} results."
+        other: "Showing top %{count} results."
       query_name: "Query"
       query_description: "Description"
       query_time: "Last run"
--- a/assets/javascripts/discourse/templates/explorer-query-result.hbs
+++ b/assets/javascripts/discourse/templates/explorer-query-result.hbs
@@ -4,6 +4,7 @@
 </div>
 
 <div class="result-about">
+  {{resultCount}}
   {{duration}}
 </div>
 
--- a/plugin.rb
+++ b/plugin.rb
@@ -939,7 +939,7 @@ SQL
 
     def create
       # guardian.ensure_can_create_explorer_query!
-      
+
       query = DataExplorer::Query.from_hash params.require(:query)
       query.created_at = Time.now
       query.created_by = current_user.id.to_s
@@ -1050,6 +1050,7 @@ SQL
               success: true,
               errors: [],
               duration: (result[:duration_secs].to_f * 1000).round(1),
+              result_count: pg_result.values.length || 0,
               params: query_params,
               columns: cols,
             }
--- a/plugin.rb
+++ b/plugin.rb
@@ -99,7 +99,7 @@ after_initialize do
 WITH query AS (
 #{query.sql}
 ) SELECT * FROM query
-LIMIT #{opts[:limit] || 250}
+LIMIT #{opts[:limit] || 1000}
 SQL
 
           time_start = Time.now


assets/javascripts/discourse/components/query-result.js.es6
```diff
@@ -35,6 +35,16 @@ const QueryResultComponent = Ember.Component.extend({
     return this.get('content.columns').length;
   }.property('content.columns.length'),
 
+  resultCount: function() {
+    if (this.get('content.result_count') === 1) {
```
rishabhnambiar: Helllo @SamSaffron :wave: 
This logic works correctly but @techAPJ mentioned that a new Data Explorer SiteSetting (maybe hidden) for result limit could be useful.
Currently, the 250-limit is hardcoded in two files. (plugin.rb, query-result.js) 
What do you think is better?

assets/javascripts/discourse/components/query-result.js.es6
```diff
@@ -35,6 +35,16 @@ const QueryResultComponent = Ember.Component.extend({
...
+      return I18n.t('explorer.single_result_count', {value: this.get('content.result_count')});
+    } else if (this.get('content.result_count') === 250) {
+      return I18n.t('explorer.result_limit_reached', {value: this.get('content.result_count')});
+    } else {
+      return I18n.t('explorer.total_result_count', {value: this.get('content.result_count')});
+    }
+  }.property('content.resultCount'),
```
techAPJ: We should use computed decorator here. For reference see: https://github.com/discourse/discourse-push-notifications/blob/2f13c67a6128a2aae8548de2eeb15dd69bfc0618/assets/javascripts/discourse/components/push-notification-config.js.es6#L8-L11

rishabhnambiar: Done, thanks! Merging...

techAPJ: wait

techAPJ: @rishabhnambiar refer this code for correct implementation: https://github.com/discourse/discourse-canned-replies/blob/303f9525105b54c8553bb375fec91b07cb01cf2c/assets/javascripts/discourse/controllers/edit-reply.js.es6#L13-L17

rishabhnambiar: I hope it's correct now, thanks so much for showing me this pattern, it's pretty awesome! :100:

config/locales/client.en.yml
```diff
@@ -52,6 +52,9 @@ en:
       download_csv: "CSV"
       others_dirty: "A query has unsaved changes that will be lost if you navigate away."
       run_time: "Query completed in {{value}} ms."
+      single_result_count: "{{value}} result."
+      total_result_count: "{{value}} results."
```
gschlager: You should use the special `%{count}` variable when you don't know the exact number. See https://guides.rubyonrails.org/i18n.html#pluralization for more on that.

I'd suggest you use the following YML instead:
```yml
result_count:
  one: "%{count} result."
  other: "%{count} results."
```

config/locales/client.en.yml
```diff
@@ -52,6 +52,9 @@ en:
...
+      result_limit_reached: "Showing top {{value}} results."
```
gschlager: This text is hard to translate, because translators don't know which number `{{value}}` will be. There are multiple ways to solve this:

1. Pluralize the string with `one` and `other`, even though the number will always stay the same.
2. Replace `{{value}}` with the actual number 250.
3. Add a comment in the line above the translation.

I'd suggest you go with 1. It might look ridiculous in English and many other languages, but it's the best way to allow translation in every language.

assets/javascripts/discourse/components/query-result.js.es6
```diff
@@ -29,8 +30,19 @@ const QueryResultComponent = Ember.Component.extend({
...
   params: Em.computed.alias('content.params'),
   explainText: Em.computed.alias('content.explain'),
-
   hasExplain: Em.computed.notEmpty('content.explain'),
+
+  @computed('content.result_count')
+  resultCount: function(count) {
+    if (count === 1) {
+      return I18n.t('explorer.single_result_count', {value: count});
+    } else if (count === 250) {
```
gschlager: Where is this magic number coming from?

rishabhnambiar: It comes from here :face_with_head_bandage:

https://github.com/discourse/discourse-data-explorer/blob/68059b540ee3af199e4f6aca0beca19e8597a350/plugin.rb#L102

assets/javascripts/discourse/components/query-result.js.es6
```diff
@@ -29,8 +30,19 @@ const QueryResultComponent = Ember.Component.extend({
   columns: Em.computed.alias('content.columns'),
   params: Em.computed.alias('content.params'),
   explainText: Em.computed.alias('content.explain'),
-
   hasExplain: Em.computed.notEmpty('content.explain'),
+
+  @computed('content.result_count')
+  resultCount: function(count) {
+    if (count === 1) {
+      return I18n.t('explorer.single_result_count', {value: count});
```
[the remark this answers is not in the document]
rishabhnambiar: I know the braces look odd but removing them seems to blow everything up :sweat_smile: 

`Oops! That page doesn’t exist or is private.`
`Error: MiniRacer::RuntimeError: SyntaxError: javascripts/discourse/components/query-result: Unexpected token, expected , (38:54)`

config/locales/client.en.yml
rishabhnambiar: > 1. Pluralize the string with one and other, even though the number will always stay the same.

That sounds great, I've done this now but is this YAML acceptable? Or should I find a way to make it work with just `one` and `other`?
```yml
result_count:
  one: "%{count} result."
  other: "%{count} results."
  max: "Showing top %{count} results."
```

I need to say `top 250 returned` so that it's clear that all results have not been displayed.
This has been discussed here: https://meta.discourse.org/t/have-data-explorer-show-number-of-results/96687/2?u=rishabhn

SamSaffron:
lets bump the max number to 1000 while we are here.

config/locales/client.en.yml
```diff
@@ -52,6 +52,10 @@ en:
       download_csv: "CSV"
       others_dirty: "A query has unsaved changes that will be lost if you navigate away."
       run_time: "Query completed in {{value}} ms."
+      result_count:
+        one: "%{count} result."
+        other: "%{count} results."
+        max: "Showing top %{count} results."
```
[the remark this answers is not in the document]
rishabhnambiar: > I know that one doesn't make much sense in English, but it's important for other languages.

I've understood now, thanks to your explanation and the i18n docs. :grin:

assets/javascripts/discourse/components/query-result.js.es6
```diff
@@ -29,8 +30,19 @@ const QueryResultComponent = Ember.Component.extend({
...
+  @computed('content.result_count')
+  resultCount: function(count) {
+    if (count === 1) {
+      return I18n.t('explorer.result_count.one', {count: count});
+    } else if (count === 250) {
+      return I18n.t('explorer.result_count.max', {count: count});
+    } else {
+      return I18n.t('explorer.result_count.other', {count: count});
+    }
+  },
```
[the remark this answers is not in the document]
rishabhnambiar: Understood! :+1: 

One small note, removing the braces still doesn't work for me.
I did a quick search in the discourse codebase where I found that only `{ count }` or `{ count: count }` was used in js files. I think the braces can be skipped only in ruby files.

PR #26, opened 2018-09-28
FEATURE: Ship default queries with the Data Explorer 

rishabhnambiar:
**How it works:** 
-Queries are added to the Data Explorer from json
-These default queries are saved in the db IF they are run
-If changes are made to the json file, the new values are updated whenever we run a query.

- [x] Reads default queries from json and adds to serializer
- [x] Hides edit/delete buttons for default queries 
- [x] Gives a persistent id to each default query
- [x] Updates last run at for default queries 
- [x] Choose which default queries to ship and make a post on meta for selection


assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
```diff
@@ -33,6 +34,7 @@ export default Ember.Controller.extend({
     const id = parseInt(this.get('selectedQueryId'));
     const item = this.get('content').find(q => q.get('id') === id);
     !isNaN(id) ? this.set('showRecentQueries', false) : this.set('showRecentQueries', true);
+    if (id<0) this.set('editDisabled', true);
```
ZogStriP: Did you run the linter on this file? Seems like this wouldn't pass.

rishabhnambiar: Prettier wants to obliterate this file :sweat_smile: Can I fix all the linting/prettier issues in a separate PR?

(I did fix that line though^)

techAPJ:
@rishabhnambiar PR looks good to me, but..

> These default queries are saved in the db IF they are run

I am not sure if we should save default query in database (on run) as any change we make to that query will not be updated on Discourse instances that had run that query.

Let's wait for @SamSaffron's opinion here.

rishabhnambiar:
Done! 
The changes now are now read and saved on every run. :+1:

SamSaffron:
This is minor, but totally worth changing. Instead of using JSON and parsing a JSON file for the queries use Ruby, it is far easier on the eye: 


```
query = Query.new(
  id: -1,
  name: "top 50 quality users",
  description: "some awesome description"
)

query.sql = <<~SQL
    SELECT something
    FROM table
SQL

@queries << query

```

plenty of ways to play with this, but the newline story in JSON is really not good at all ... makes it hard to review and add queries

techAPJ:
@rishabhnambiar while you are at it make sure to squash all commits into a single commit for better git commit history.

lib/queries.rb
```diff
@@ -0,0 +1,275 @@
+class Queries
+  def self.default
+    queries = {
```
SamSaffron: quries{
  :common_likers = {
      id: -1, name: 'common likers', description: ''
   }
}

Cause then when you append the sql you have a sane name to use instead of needing to have -1 in two spots in the file

rishabhnambiar: I agree @SamSaffron, this is very very ugly but I'm simply accessing the hash by id in `def find()` and in def `save_default_query()`, depending on the first key to be the id. I need to be able to search the queries to be able to run it so I'm finding by key here.

plugin.rb
```diff
@@ -630,8 +631,7 @@ def to_hash
 
     def self.find(id, opts = {})
       if DataExplorer.pstore_get("q:#{id}").nil? && id < 0
-        json = JSON.parse(File.read(File.expand_path("../config/default_queries.json", __FILE__)))
-        hash = json["queries"][id.to_s].with_indifferent_access
+        hash = Queries.default[id.to_s]
```
rishabhnambiar: @SamSaffron, simply indexing by id here and below. Should I rework this to find by value and not key?

SamSaffron: At the end of the file you can rework your query hash to index by id, or allow id to be text which is kind of nice

lib/queries.rb
```diff
@@ -0,0 +1,280 @@
...
+        "mostcommonlikers": {
```
SamSaffron: lets clean this up ... I like `user_participation` as a pattern... I think it is the right direction... maybe

`most-messages`, `top-likers`, `new-topics` ... looks kind of pretty in the url that way.

rishabhnambiar: Oh yes, I've cleaned this up now. It is so much better to look at when editing the sql :+1:

PR #28, opened 2018-11-01
FEATURE: Replace combo-box with search filter

rishabhnambiar:
---

### DIFF
--- a/assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
+++ b/assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
@@ -2,6 +2,7 @@ import showModal from "discourse/lib/show-modal";
 import Query from "discourse/plugins/discourse-data-explorer/discourse/models/query";
 import { popupAjaxError } from "discourse/lib/ajax-error";
 import { ajax } from "discourse/lib/ajax";
+import { default as computed } from "ember-addons/ember-computed-decorators";
 
 const NoQuery = Query.create({ name: "No queries", fake: true });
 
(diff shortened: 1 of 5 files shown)


assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
```diff
@@ -26,6 +26,16 @@ export default Ember.Controller.extend({
   sortBy: ["last_run_at:desc"],
   sortedQueries: Em.computed.sort("model", "sortBy"),
 
+  searchedContent: function() {
```
techAPJ: We should use `computed` along with ES6 function syntax here.

assets/javascripts/discourse/templates/admin/plugins-explorer.hbs
```diff
@@ -164,7 +164,7 @@
           </thead>
           <tr></tr>
           <tbody>
-            {{#each sortedQueries as |query|}}
+            {{#each searchedContent as |query|}}
```
techAPJ: Nitpicking.. but I think `filteredContent` looks better here?

techAPJ:
Added a comment to use new syntax, other than that looks good. 👍

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
rishabhnambiar: Done, sorry!

assets/javascripts/discourse/templates/admin/plugins-explorer.hbs
rishabhnambiar: Definitely, updated :+1:

PR #32, opened 2019-03-30
WIP: Add Group Members Reply Count query

scossar:


scossar:
I'm going to make a couple of small changes to the query:

 I don't think the `months_ago` parameter needs to be used here. I'll change it so that the query period is set with `start_date` and `end_date` parameters.

The `target_users` query isn't excluding users who were not members of the group during the query period. I'll fix it so that results are only returned for users who belonged to the group during the query period.

scossar:
I've submitted a new PR instead of rebasing this one.

PR #36, opened 2019-08-29
Allow groups to access queries

markvanlan:
The thread describing this feature can be found [here](https://meta.discourse.org/t/make-data-explorer-available-to-moderators).


assets/javascripts/discourse/connectors/group-reports-nav-item/nav-item.js.es6
```diff
@@ -0,0 +1,14 @@
+import { ajax } from "discourse/lib/ajax";
+
+export default {
+  shouldRender(args, component) {
+    let usersGroupIds = component.currentUser.groups.map(g => g.id)
+    if (!usersGroupIds.includes(args.groupId)) return false // User not a part of group
+
+    return true
```
SamSaffron: be sure to install prettier locally and always run your files through it. I use IDE integration it is magical.

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
```diff
@@ -81,6 +100,10 @@ export default Ember.Controller.extend({
     this.set("loading", true);
     if (this.get("selectedItem.description") === "")
       this.set("selectedItem.description", "");
+    // params not comin in
```
SamSaffron: comment a bit confusing

assets/javascripts/discourse/routes/admin-plugins-explorer.js.es6
```diff
@@ -6,17 +6,21 @@ export default Discourse.Route.extend({
   model() {
     const p1 = this.store.findAll("query");
     const p2 = ajax("/admin/plugins/explorer/schema.json", { cache: true });
+    const p3 = ajax("/admin/plugins/explorer/groups.json"); // Should cache be true?
```
SamSaffron: This will not cache ... its not really a cachable route, one consideration here is preload store, may make sense to use it. or amend serializer to include groups

lib/queries.rb
```diff
@@ -14,77 +14,92 @@ def self.default
         "most-common-likers": {
             "id": -1,
             "name": "Most Common Likers",
-            "description": "Which users like particular other users the most?"
+            "description": "Which users like particular other users the most?",
+            "group_ids": []
```
SamSaffron: if it is just going to be blank everywhere is there any point including it?

plugin.rb
```diff
@@ -1048,6 +1055,10 @@ def schema
       end
     end
 
+    def groups
+      render json: Group.select(:id, :name)
```
SamSaffron: This bypasses security, probably not mega risky as it is admin only, but why not rely on existing group api?

lib/queries.rb
markvanlan: Yeah, agreed.

plugin.rb
markvanlan: Agreed, no reason to not use existing group api.

assets/javascripts/discourse/routes/admin-plugins-explorer.js.es6
markvanlan: Using existing group api now

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
```diff
@@ -81,6 +97,15 @@ export default Ember.Controller.extend({
     this.set("loading", true);
```
SamSaffron: I think a lot of this can be amended to stop all this `.get` and `.set`, cause we this.selectedItem is already assumed to be there in the return. So `this.selectedItem.group_ids && this.selectedItem.group_ids.length < 1` and so on in this method make sense.

assets/javascripts/discourse/controllers/group-reports-show.js.es6
```diff
@@ -0,0 +1,46 @@
...
+  results: Ember.computed.alias("model.results"),
+  hasParams: Ember.computed.gt("model.param_info.length", 0),
+
+  actions: {
+    run() {
+      this.setProperties({ loading: true, showResults: false });
+      ajax(`/g/${this.get("group.name")}/reports/${this.get("model.id")}/run`, {
+        type: "POST",
+        data: {
+          params: JSON.stringify(this.get("model.params")),
```
SamSaffron: isn't model a given? this.model.id?

assets/javascripts/discourse/controllers/group-reports-show.js.es6
```diff
@@ -0,0 +1,46 @@
...
+        }
+      })
+        .then(result => {
+          this.set("results", result);
+          if (!result.success) {
+            this.set("showResults", false);
+            return;
+          }
+
+          this.set("showResults", true);
```
SamSaffron: a lot of `set` here, I don't think we need this anymore `this.showResults = false` should do

spec/controllers/queries_controller_spec.rb
```diff
@@ -13,238 +11,362 @@ def response_json
...
-    it "can export data in CSV format" do
-      q = make_query('SELECT 23 as my_value')
-      post :run, params: { id: q.id, download: 1 }, format: :csv
-      expect(response.status).to eq(200)
+  describe "#user_can_access_query" do
+    let(:group) { Fabricate(:group) }
+
+    before do
+      controller
+        .stubs(:params)
```
SamSaffron: I much prefer no stubs here, can this be avoided?

SamSaffron:
Overall this looks great, I also love the UX and positioning! great job

General pattern though it is a bit hard for me to review every line of the spec file, but can you confirm the new public controller actions all have an integration test to ensure users who are not allowed have no access and one's allowed do.

SamSaffron:
test

SamSaffron:
@eviltrout can you have a look as well?

assets/javascripts/discourse/components/group-reports-nav-item.js.es6
```diff
@@ -0,0 +1,23 @@
+import { ajax } from "discourse/lib/ajax";
+
+export default Ember.Component.extend({
+  group: null,
+  showReportsTab: false,
+
+  checkForReports() {
+    const p1 = ajax(`/g/${this.group.name}/reports`);
```
eviltrout: Trivial but you could avoid the `p1` variable here by returning `ajax(...).then()`

assets/javascripts/discourse/components/group-reports-nav-item.js.es6
```diff
@@ -0,0 +1,23 @@
...
+    return p1.then(response => {
+      return this.set("showReportsTab", response.queries.length > 0);
+    });
+  },
+
+  init(args) {
+    this.set("group", args.group);
+    let usersGroupIds = this.currentUser.groups.map(g => g.id);
```
eviltrout: Rather than mapping through then running an `includes`, you could do it in one go with something like: `if (this.currentUser.groups.some(g => g.id === this.group.id)`

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
```diff
@@ -81,6 +97,15 @@ export default Ember.Controller.extend({
...
     if (this.get("selectedItem.description") === "")
       this.set("selectedItem.description", "");
+
+    // group_ids must not be an empty array, as the param will not be sent the server.
```
eviltrout: I don't like using the same variable to represent two different actions. Rather than using a magic `-1` value to represent this I would far prefer if you added a second parameter that indicates "remove all".

assets/javascripts/discourse/controllers/group-reports-index.js.es6
```diff
@@ -0,0 +1,10 @@
+import Query from "discourse/plugins/discourse-data-explorer/discourse/models/query";
+import { ajax } from "discourse/lib/ajax";
+import {
+	default as computed,
+	observes
+} from "ember-addons/ember-computed-decorators";
+
+export default Ember.Controller.extend({
+	queries: Ember.computed.alias("model.queries")
```
eviltrout: The indentation here is odd. Should be two spaces.

assets/javascripts/discourse/routes/admin-plugins-explorer.js.es6
```diff
@@ -4,19 +4,36 @@ export default Discourse.Route.extend({
...
     const p2 = ajax("/admin/plugins/explorer/schema.json", { cache: true });
-    return p1
-      .then(model => {
-        model.forEach(query => query.markNotDirty());
+    const p3 = this.store.findAll("query");
 
+    return p1
+      .then(groups => {
+        let group_names = {};
+        groups.forEach(g => {
```
eviltrout: Another trivial suggestion but this could be one line. `groups.forEach(g => groupNames[g.id] = g.name)`

assets/javascripts/discourse/routes/admin-plugins-explorer.js.es6
```diff
@@ -4,19 +4,36 @@ export default Discourse.Route.extend({
...
+    const p1 = this.store.findAll("group");
     const p2 = ajax("/admin/plugins/explorer/schema.json", { cache: true });
-    return p1
-      .then(model => {
-        model.forEach(query => query.markNotDirty());
+    const p3 = this.store.findAll("query");
 
+    return p1
+      .then(groups => {
+        let group_names = {};
```
eviltrout: In Javascript code we use `camelCase` not `snake_case` for variables.

assets/javascripts/discourse/templates/admin/plugins-explorer.hbs
```diff
@@ -52,6 +53,20 @@
               </div>
             {{/if}}
 
+            <div class="groups">
+              <span class="label">Allow groups to acess this query</span>
+              <span>
+                {{ multi-select values=selectedItem.group_ids content=groupOptions }}
```
eviltrout:
```suggestion
                {{multi-select values=selectedItem.group_ids content=groupOptions}}
```

assets/javascripts/discourse/templates/admin/plugins-explorer.hbs
```diff
@@ -52,6 +53,20 @@
...
+              </span>
+              {{#if runDisabled }}
```
eviltrout:
```suggestion
              {{#if runDisabled}}
```

assets/javascripts/discourse/templates/components/group-reports-nav-item.hbs
```diff
@@ -0,0 +1,9 @@
+{{#if showReportsTab}}
+	<ul class ='nav-pills'>
```
eviltrout: Again quite weird indentation here.

assets/javascripts/discourse/templates/connectors/group-reports-nav-item/nav-item.hbs
```diff
@@ -0,0 +1 @@
+{{ group-reports-nav-item group = this.group }}
```
eviltrout:
```suggestion
{{group-reports-nav-item group = this.group}}
```

eviltrout: Does this work properly? You shouldn't need to use `this` in a hbs template.

assets/javascripts/discourse/templates/group-reports-index.hbs
```diff
@@ -0,0 +1,25 @@
...
+      </th>
+      <th>
+        {{i18n "explorer.query_time"}}
+      </th>
+    </thead>
+    <tr></tr>
+    <tbody>
+      {{#each queries as |query|}}
+        <tr>
+          <td><a href="reports/{{query.id}}">{{query.name}}</a></td>
```
eviltrout: Can you use `{{link-to}}` here? If not this link will need to be aware of subfolder installations of Discourse. (usually a call to `Discourse.getURL`

plugin.rb
```diff
@@ -996,6 +1019,39 @@ def show
       render_serialized query, DataExplorer::QuerySerializer, root: 'query'
     end
 
+    def group_reports_index
+      respond_to do |format|
+        format.html { render 'groups/show' }
+        format.json do        
+          group = Group.find_by(name: params["group_name"])
+          render status: 404, json: {} and return unless user_included_in_group(group)
```
eviltrout:
```suggestion
          return render(status: 404, json: {}) unless user_included_in_group(group)
```

plugin.rb
```diff
@@ -996,6 +1019,39 @@ def show
...
+        end
+      end
+    end
+
+    def group_reports_show
+      respond_to do |format|
+        format.html { render 'groups/show' }
+        format.json do    
+          query = DataExplorer::Query.find(params[:id].to_i)  
+          render status: 404, json: {} and return unless user_can_access_query(query)
```
eviltrout:
```suggestion
          return render(status: 404, json: {}) unless user_can_access_query(query)
```

assets/javascripts/discourse/components/group-reports-nav-item.js.es6
markvanlan: Agreed!

assets/javascripts/discourse/components/group-reports-nav-item.js.es6
markvanlan: 👍

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
markvanlan: Great idea. I will do that.

assets/javascripts/discourse/templates/connectors/group-reports-nav-item/nav-item.hbs
markvanlan: It does work, but I can certainly remove `this`.

assets/javascripts/discourse/routes/admin-plugins-explorer.js.es6
markvanlan: @eviltrout Prettier is forcing it onto multiple line

assets/javascripts/discourse/templates/group-reports-index.hbs
markvanlan: We can!

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
markvanlan: All of this is unnecessary, and has been removed.

spec/controllers/queries_controller_spec.rb
markvanlan: > Overall this looks great, I also love the UX and positioning! great job
> 
> General pattern though it is a bit hard for me to review every line of the spec file, but can you confirm the new public controller actions all have an integration test to ensure users who are not allowed have no access and one's allowed do.

@SamSaffron I have added integration tests (before I had filters, and tested those, but never actually hitting the endpoint that use the filters), and removed the stubbing.

markvanlan:
> Thanks for the big piece of work. I've provided feedback.

@eviltrout I addressed all your comments. Thanks for the thorough review!

assets/javascripts/discourse/components/group-reports-nav-item.js.es6
```diff
@@ -0,0 +1,21 @@
+import { ajax } from "discourse/lib/ajax";
+
+export default Ember.Component.extend({
+  group: null,
+  showReportsTab: false,
+
+  checkForReports() {
+    return ajax(`/g/${this.group.name}/reports`).then(response => {
+      return this.set("showReportsTab", response.queries.length > 0);
```
SamSaffron: There are lots of places in this PR that use `.set` no need really.

`this.showReportsTab = response.queries.length > 0;` reads much better (there are quite a few similar places in the PR)

assets/javascripts/discourse/routes/admin-plugins-explorer.js.es6
```diff
@@ -4,19 +4,36 @@ export default Discourse.Route.extend({
   controllerName: "admin-plugins-explorer",
 
   model() {
-    const p1 = this.store.findAll("query");
```
SamSaffron: probably worth renaming this to queryPromise and so on ... p1,p2,p3 is somewhat confusing.

assets/javascripts/discourse/routes/group-reports-index.js.es6
```diff
@@ -0,0 +1,39 @@
+import { ajax } from "discourse/lib/ajax";
+
+export default Discourse.Route.extend({
+  controllerName: "group-reports-index",
+
+  model() {
+    const group = this.modelFor("group");
+    const p1 = ajax(`/g/${group.name}/reports`);
+    return p1
```
SamSaffron: no need to define p1 here, `return ajax(...`

plugin.rb
```diff
@@ -996,6 +1019,39 @@ def show
...
+          return render(status: 404, json: {}) unless current_user && user_can_access_query(query)
+
+          render_serialized query, DataExplorer::QuerySerializer, root: 'query'
+        end
+      end
+    end
+
+    def group_reports_run
+      query = DataExplorer::Query.find(params[:id].to_i)
+      return render(status: 404, json: {}) unless current_user && user_can_access_query(query)
```
SamSaffron: Our general pattern is to use a guardian eg: that should take care of current_user check and everything. 

https://github.com/discourse/discourse/blob/cd894b7d17a5d0712facb943229cdb2504818a7f/app/controllers/topics_controller.rb#L49-L49

plugin.rb
```diff
@@ -996,6 +1019,39 @@ def show
...
+        end
+      end
+    end
+
+    def group_reports_show
+      respond_to do |format|
+        format.html { render 'groups/show' }
+        format.json do    
+          query = DataExplorer::Query.find(params[:id].to_i)
+          return render(status: 404, json: {}) unless current_user && user_can_access_query(query)
```
SamSaffron: a guardian check at the entry is what we want here, no to scope it down to json only. the html should 404 as well.

assets/javascripts/discourse/components/group-reports-nav-item.js.es6
markvanlan: @SamSaffron I agree it is cleaner, but I kept getting errors when trying to set without calling `this.set`. I got this error when I tried to not use set.

image
image

SamSaffron: :cry:  I wonder if there is a workaround here for components @eviltrout / @jjaffeux  ? Seems like a tricky rule to remember.

discoursereviewbot: [Joffrey JAFFEUX posted](https://review.discourse.org/t/allow-groups-to-access-queries-pr-36/5571/46):

AFAIK octane should mostly land in 3.14 https://blog.emberjs.com/2019/08/15/octane-release-plan.html and this is when we should start to have solutions to this.

Can’t find it in this blog post post, but I remember reading something with `@tracked` which should make this super explicit.

assets/javascripts/discourse/components/group-reports-nav-item.js.es6
```diff
@@ -0,0 +1,21 @@
...
+    });
+  },
+
+  init(args) {
+    this.set("group", args.group);
+    if (this.currentUser.groups.some(g => g.id === this.group.id)) {
+      // User is apart of the group. Now check if the group has reports
```
eviltrout:
```suggestion
      // User is a part of the group. Now check if the group has reports
```

SamSaffron:
Looking good, @eviltrout  probably best if you merge this in your morning.

eviltrout:
Merged, let's do it!

PR #37, opened 2019-09-16
Fix: ability to import an exported query

markvanlan:
`this.files` was undefined, and we already have access to the file input, so now calling that directly.


SamSaffron:
I think this looks right

PR #41, opened 2019-10-24
FEATURE: Popup to copy shareable links to reports

markvanlan:
[Thread](https://meta.discourse.org/t/data-explorer-allow-groups-to-access-not-working/131595/5?u=markvanlan) about the issue.

### DIFF
--- a/assets/javascripts/discourse/components/share-report.js.es6
+++ b/assets/javascripts/discourse/components/share-report.js.es6
@@ -2,8 +2,6 @@ import {
   default as computed,
   on
 } from "ember-addons/ember-computed-decorators";
-import Sharing from "discourse/lib/sharing";
-import { nativeShare } from "discourse/lib/pwa-utils";
 
 export default Ember.Component.extend({
   classNames: ["share-report"],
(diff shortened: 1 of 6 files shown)


assets/javascripts/discourse/components/share-report.js.es6
```diff
@@ -0,0 +1,88 @@
...
+  classNames: ["share-report"],
+
+  group: null,
+  query: null,
+  visible: false,
+
+  @computed("group", "query")
+  link() {
+    return (
+      window.location.protocol +
```
eviltrout: Can we use `Discourse.BaseUrl` here? It might also look nicer using template strings instead of concatenation.

assets/javascripts/discourse/templates/components/share-report.hbs
```diff
@@ -0,0 +1,13 @@
+<div onclick={{action "open"}}>
+  {{d-icon "link"}}
+  {{group}}
+</div>
+{{#if visible}}
+  <div class='popup'>
+    <label>Link for {{group}}</label>
```
eviltrout: Missing i18n here

PR #45, opened 2019-12-07
FEATURE: Additional enums

tshenry:
This adds several new enums to help Data Explorer users quickly reference the meaning of the more obscure column values. I also removed `groups.alias_level': Group::ALIAS_LEVELS` since it is not actually a column listed in the Data Explorer. Instead, I split it into `groups.mentionable_level': Group::ALIAS_LEVELS` and `groups.messageable_level': Group::ALIAS_LEVELS`.

In testing, I noticed that there are some rather large values that break the design. @awesomerobot might want to follow this up with a design pass. My initial thought was to widen the problem areas:

#### Before:
Screen Shot 2019-12-06 at 4 21 11 PM

#### After:
Screen Shot 2019-12-06 at 4 18 51 PM

But I'm not 100% sure if that's the best approach here, so I held off on adding any styling changes in this PR.


SamSaffron:
After looks a lot better to me than before, I say go with it at least temporarily.

tshenry:
> After looks a lot better to me than before, I say go with it at least temporarily.

Okay, I followed this up with the following commit for now: https://github.com/discourse/discourse-data-explorer/commit/02f4277f3dfa31bd39ed70f5cfd0491cf18d6e8c

I'll also give Kris a heads up in case he wants to do anything further.

PR #49, opened 2020-05-19
UX: minor share report popup improvements

tshenry:
This change will make sure color schemes are honored, add `type="text"` to the `<input>` element, and reuse the "close" icon button from core's share-link popup.

**Before:**
Screen Shot 2020-05-18 at 4 26 16 PM

**After:**
Screen Shot 2020-05-18 at 4 26 27 PM

### DIFF
--- a/assets/javascripts/discourse/templates/components/share-report.hbs
+++ b/assets/javascripts/discourse/templates/components/share-report.hbs
@@ -6,6 +6,6 @@
   <div class='popup'>
     <label>{{i18n "explorer.link"}} {{group}}</label>
     <input type="text" value={{link}}/>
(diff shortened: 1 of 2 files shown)


assets/javascripts/discourse/templates/components/share-report.hbs
```diff
@@ -5,9 +5,7 @@
 {{#if visible}}
   <div class='popup'>
     <label>{{i18n "explorer.link"}} {{group}}</label>
-    <input value={{link}}/>
-    <span onclick={{action "close"}}>
-      {{d-icon "times"}}
-    </span>
+    <input type="text" value={{link}}/>
+    {{d-button action="close" class="btn btn-flat close" icon="times" aria-label="share.close" title="share.close"}}
```
jjaffeux, line 9: no need for "btn" in class here, it's already present with d-button

PR #51, opened 2020-06-24
PERF: Stream CSV response bodies

riking:
Additionally, restore 1 exponent to the result limit for CSV downloads.

This still does not stream results from pg->web, but is still worth it because this ~halves the peak memory use for large queries.


riking:
DO NOT MERGE until release is settled

riking:
Doesn't actually work. We're going to need to `rack.hijack`.

PR #52, opened 2020-07-06
Import getURL to fix deprecation warnings

markvanlan:


SamSaffron:
odd ... how has this survived so long :) ... merging it in

PR #53, opened 2020-07-11
FEATURE: Add ability to toggle visibility for individual queries

RickyC0626:
- User can click the hide/show button to toggle the query's visibility
- When the Hide/Show Hidden Queries button is toggled, it will
    - Show hidden queries with gray background when on
    - Hide hidden queries when off
- Visibility state of each query is persisted upon page refresh

discourse-data-explorer-hide


RickyC0626:
See #54 for updated implementation

PR #58, opened 2020-08-10
UI: Fixed missing text above query list for sorting

AndreaHabib:
* changed directory-toggle to table-header-toggle which fixed the issue


RickyC0626:
Future proofing for deprecation of [directory-toggle](https://github.com/discourse/discourse/blob/master/app/assets/javascripts/discourse/app/components/directory-toggle.js) in Discourse versions above 2.5

markvanlan:
Looks great, thanks!

PR #59, opened 2020-08-11
UI: Don't show 'everyone' group for query access

RickyC0626:
[Query access for 'everyone' does not work, but trust level 0 does, and it works the same](https://meta.discourse.org/t/allow-everyone-access-to-data-explorer-query-doesnt-work/160292)

**Before:**
'everyone' group is listed
image

**After:**
'everyone' group not listed anymore
image


AndreaHabib:
Errors that were caused by this issue:
Error: Can’t hydrate group without primaryKey: id

SamSaffron:
This change looks safe to me! thanks @RickyC0626

SamSaffron (approved):

PR #60, opened 2020-08-17
FIX: Use updated routes to link user profiles

rishabhnambiar:
This fixes links for both, the result view:

image

and the query view:

image

### DIFF
--- a/assets/javascripts/discourse/templates/admin/plugins-explorer.hbs
+++ b/assets/javascripts/discourse/templates/admin/plugins-explorer.hbs
@@ -201,7 +201,7 @@
                 </td>
                 <td class="query-created-by">
                   {{#if query.username}}
-                    <a href="/users/{{query.username}}/activity">
+                    <a href="/u/{{query.username}}/activity">
                       <medium>{{query.username}}</medium>
                     </a>
                   {{/if}}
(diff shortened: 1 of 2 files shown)


assets/javascripts/discourse/templates/admin/plugins-explorer.hbs
```diff
@@ -201,7 +201,7 @@
                 </td>
                 <td class="query-created-by">
                   {{#if query.username}}
-                    <a href="/users/{{query.username}}/activity">
+                    <a href="/u/{{query.username}}/activity">
```
ZogStriP, line 204: Does this work when using `subfolder`?

rishabhnambiar: Are you asking if the fix works or the route? The route works as you can see on https://www.livechat.com/community/u/system/summary.

ZogStriP: Sure, but I don't think that URL is the right one when using a `subfolder`. It's missing the `baseUri` like in the raw hbs file.

rishabhnambiar: I tested this on a subfolder site and both of these links lead to `https://www.example.com/forum/users/system/activity`. When I manually change the `users` to `u` in the url, the link then works without any other changes. So I think the sub-folder part before the suffix is wired up correctly. 

Does that sound right?

PR #61, opened 2020-08-21
FEATURE: introduce data-explorer tables

lis2:
Instead of using `PluginStoreRow` we should use plugin-specific models like `DataExplorer::Query` and `DataExplorer::QueryGroup`

@tgxworld I didn't create that controller, I was just thinking that extracting that to separate file would make sense. However, I had to make small adjustments to use separate models instead of PluginStoreRows

### DIFF
--- a/app/controllers/data_explorer/query_controller.rb
+++ b/app/controllers/data_explorer/query_controller.rb
@@ -1,97 +1,78 @@
 class DataExplorer::QueryController < ::ApplicationController
   requires_plugin DataExplorer.plugin_name
 
-  before_action :check_enabled
   before_action :set_group, only: %i(group_reports_index group_reports_show group_reports_run)
-  before_action :set_query, only: %i(group_reports_show group_reports_run)
-
-  attr_reader :group, :query
-
-  def check_enabled
-    raise Discourse::NotFound unless SiteSetting.data_explorer_enabled?
-  end
-
-  def set_group
-    @group = Group.find_by(name: params["group_name"])
-  end
-
-  def set_query
-    @query = DataExplorer::Query.find(params[:id])
-  end
+  before_action :set_query, only: %i(group_reports_show group_reports_run show update)
+  skip_before_action :check_xhr, only: %i(show group_reports_run run)
 
   def index
-    queries = DataExplorer::Query.where(hidden: false).order(:last_run_at, :name).to_a
+    queries = DataExplorer::Query.where(hidden: false).order(:last_run_at, :name).includes(:groups).to_a
 
     database_queries_ids = DataExplorer::Query.pluck(:id)
     Queries.default.each do |params|
       attributes = params.last
+      next if database_queries_ids.include?(attributes["id"])
       query = DataExplorer::Query.new
       query.id = attributes["id"]
       query.sql = attributes["sql"]
       query.name = attributes["name"]
       query.description = attributes["description"]
       query.user_id = Discourse::SYSTEM_USER_ID.to_s
-      queries << query unless database_queries_ids.include?(query.id)
+      queries << query 
     end
 
     render_serialized queries, DataExplorer::QuerySerializer, root: 'queries'
   end
 
-  skip_before_action :check_xhr, only: [:show]
   def show
     check_xhr unless params[:export]
 
-    query = DataExplorer::Query.find(params[:id])
-
     if params[:export]
-      response.headers['Content-Disposition'] = "attachment; filename=#{query.slug}.dcquery.json"
+      response.headers['Content-Disposition'] = "attachment; filename=#{@query.slug}.dcquery.json"
       response.sending_file = true
     end
 
-    return raise Discourse::NotFound if !guardian.user_can_access_query?(query) || query.hidden
-    render_serialized query, DataExplorer::QuerySerializer, root: 'query'
+    return raise Discourse::NotFound if !guardian.user_can_access_query?(@query) || @query.hidden
+    render_serialized @query, DataExplorer::QuerySerializer, root: 'query'
   end
 
   def groups
-    render_serialized(Group.all, BasicGroupSerializer)
+    render json: Group.all.select(:id, :name), root: false
   end
 
   def group_reports_index
-    return raise Discourse::NotFound unless guardian.user_is_a_member_of_group?(group)
+    return raise Discourse::NotFound unless guardian.user_is_a_member_of_group?(@group)
 
     respond_to do |format|
-      format.html { render 'groups/show' }
       format.json do
         queries = DataExplorer::Query
           .where(hidden: false)
-          .joins("INNER JOIN data_explorer_query_groups data_explorer_query_groups 
+          .joins("INNER JOIN data_explorer_query_groups
                     ON data_explorer_query_groups.query_id = data_explorer_queries.id 
-                    AND data_explorer_query_groups.group_id = #{group.id}")
+                    AND data_explorer_query_groups.group_id = #{@group.id}")
         render_serialized(queries, DataExplorer::QuerySerializer, root: 'queries')
       end
     end
   end
 
   def group_reports_show
-    return raise Discourse::NotFound if !guardian.group_and_user_can_access_query?(group, query) || query.hidden
+    return raise Discourse::NotFound if !guardian.group_and_user_can_access_query?(@group, @query) || @query.hidden
 
     respond_to do |format|
-      format.html { render 'groups/show' }
       format.json do
-        render_serialized query, DataExplorer::QuerySerializer, root: 'query'
+        render_serialized @query, DataExplorer::QuerySerializer, root: 'query'
       end
     end
   end
 
-  skip_before_action :check_xhr, only: [:group_reports_run]
   def group_reports_run
-    return raise Discourse::NotFound if !guardian.group_and_user_can_access_query?(group, query) || query.hidden
+    return raise Discourse::NotFound if !guardian.group_and_user_can_access_query?(@group, @query) || @query.hidden
 
     run
   end
 
   def create
-    query = DataExplorer::Query.create!(params.require(:query).permit(:name, :description, :sql).merge(user_id: current_user.id, last_run_at: Time.now).permit!)
+    query = DataExplorer::Query.create!(params.require(:query).permit(:name, :description, :sql).merge(user_id: current_user.id, last_run_at: Time.now))
     group_ids = params.require(:query)[:group_ids]
     group_ids&.each do |group_id|
       query.query_groups.find_or_create_by!(group_id: group_id)
@@ -100,16 +81,17 @@ class DataExplorer::QueryController < ::ApplicationController
   end
 
   def update
-    query = DataExplorer::Query.find(params[:id])
-    query.update!(params.require(:query).permit(:name, :sql, :description).merge(hidden: false))
+    ActiveRecord::Base.transaction do
+      @query.update!(params.require(:query).permit(:name, :sql, :description).merge(hidden: false))
 
-    group_ids = params.require(:query)[:group_ids]
-    DataExplorer::QueryGroup.where.not(group_id: group_ids).delete_all
-    group_ids&.each do |group_id|
-      query.query_groups.find_or_create_by!(group_id: group_id)
+      group_ids = params.require(:query)[:group_ids]
+      DataExplorer::QueryGroup.where.not(group_id: group_ids).where(query_id: @query.id).delete_all
+      group_ids&.each do |group_id|
+        @query.query_groups.find_or_create_by!(group_id: group_id)
+      end
     end
 
-    render_serialized query, DataExplorer::QuerySerializer, root: 'query'
+    render_serialized @query, DataExplorer::QuerySerializer, root: 'query'
   rescue DataExplorer::ValidationError => e
     render_json_error e.message
   end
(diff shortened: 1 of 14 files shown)


app/controllers/data_explorer/query_controller.rb
```diff
@@ -0,0 +1,236 @@
+class DataExplorer::QueryController < ::ApplicationController
+  requires_plugin DataExplorer.plugin_name
+
+  before_action :check_enabled
```
tgxworld, line 4: `requires_plugin` is already setting a `before_action` to check that the plugin has not been disabled.

app/controllers/data_explorer/query_controller.rb
```diff
@@ -0,0 +1,236 @@
...
+  before_action :set_group, only: %i(group_reports_index group_reports_show group_reports_run)
+  before_action :set_query, only: %i(group_reports_show group_reports_run)
+
+  attr_reader :group, :query
```
tgxworld, line 8: :thinking: Are we using these anywhere?

app/controllers/data_explorer/query_controller.rb
```diff
@@ -0,0 +1,236 @@
...
+
+  def check_enabled
+    raise Discourse::NotFound unless SiteSetting.data_explorer_enabled?
+  end
+
+  def set_group
```
tgxworld, line 14: Let us make this method private since they are not controller actions.

app/controllers/data_explorer/query_controller.rb
```diff
@@ -0,0 +1,236 @@
...
+    @group = Group.find_by(name: params["group_name"])
+  end
+
+  def set_query
```
tgxworld, line 18: Let us make this method private since they are not controller actions.

app/controllers/data_explorer/query_controller.rb
```diff
@@ -0,0 +1,236 @@
...
+    @query = DataExplorer::Query.find(params[:id])
```
tgxworld, line 19: `find` will raise an error if the `params[:id]` is invalid. Do we want to raise an error here? Otherwise, it'll probably be better to raise a `Discourse::InvalidParameters` error or maybe `Discourse::NotFound.`

app/controllers/data_explorer/query_controller.rb
```diff
@@ -0,0 +1,236 @@
...
+    database_queries_ids = DataExplorer::Query.pluck(:id)
+    Queries.default.each do |params|
+      attributes = params.last
+      query = DataExplorer::Query.new
+      query.id = attributes["id"]
+      query.sql = attributes["sql"]
+      query.name = attributes["name"]
+      query.description = attributes["description"]
+      query.user_id = Discourse::SYSTEM_USER_ID.to_s
+      queries << query unless database_queries_ids.include?(query.id)
```
tgxworld, line 34: I think we can do this check much earlier to avoid all the work above if `database_queries_ids.include?(query.id)` returns true.

app/controllers/data_explorer/query_controller.rb
```diff
@@ -0,0 +1,236 @@
...
+    end
+
+    render_serialized queries, DataExplorer::QuerySerializer, root: 'queries'
+  end
+
+  skip_before_action :check_xhr, only: [:show]
```
tgxworld, line 40: Generally I prefer to move this to the top of the controller so that we can see all the callbacks in a single place.

app/controllers/data_explorer/query_controller.rb
```diff
@@ -0,0 +1,236 @@
...
+  def show
+    check_xhr unless params[:export]
+
+    query = DataExplorer::Query.find(params[:id])
```
tgxworld, line 44: This will raise an error if `params[:id]` is invalid.

app/controllers/data_explorer/query_controller.rb
```diff
@@ -0,0 +1,236 @@
+class DataExplorer::QueryController < ::ApplicationController
```
tgxworld, line 1: I think the plugin is only accessible to admin so we need proper access control here.

app/controllers/data_explorer/query_controller.rb
```diff
@@ -0,0 +1,236 @@
...
+      response.headers['Content-Disposition'] = "attachment; filename=#{query.slug}.dcquery.json"
+      response.sending_file = true
+    end
+
+    return raise Discourse::NotFound if !guardian.user_can_access_query?(query) || query.hidden
+    render_serialized query, DataExplorer::QuerySerializer, root: 'query'
+  end
+
+  def groups
+    render_serialized(Group.all, BasicGroupSerializer)
```
tgxworld, line 56: Just curious, how is this being used because performance might be bad if a site has alot of groups. Also, do we need to filter out automatic groups or implement some sort of access control here?

app/controllers/data_explorer/query_controller.rb
tgxworld: Ahhh looks like there is an admin constraint on the route so this is OK.

app/controllers/data_explorer/query_controller.rb
```diff
@@ -0,0 +1,236 @@
...
+  end
+
+  def group_reports_index
+    return raise Discourse::NotFound unless guardian.user_is_a_member_of_group?(group)
+
+    respond_to do |format|
+      format.html { render 'groups/show' }
```
tgxworld, line 63: I'm not quite sure why we need this, is this safe to remove?

app/controllers/data_explorer/query_controller.rb
```diff
@@ -0,0 +1,236 @@
...
+
+    run
+  end
+
+  def create
+    query = DataExplorer::Query.create!(params.require(:query).permit(:name, :description, :sql).merge(user_id: current_user.id, last_run_at: Time.now).permit!)
+    group_ids = params.require(:query)[:group_ids]
+    group_ids&.each do |group_id|
+      query.query_groups.find_or_create_by!(group_id: group_id)
+    end
```
tgxworld, line 98: We should execute all the queries in a single transaction so that one query failing does not leave us in a weird state.

app/controllers/data_explorer/query_controller.rb
```diff
@@ -0,0 +1,236 @@
...
+    render_serialized query, DataExplorer::QuerySerializer, root: 'query'
+  end
+
+  def update
+    query = DataExplorer::Query.find(params[:id])
```
tgxworld, line 103: This will raise an error on invalid `params[:id]` too.

app/controllers/data_explorer/query_controller.rb
```diff
@@ -0,0 +1,236 @@
...
+    query.update!(params.require(:query).permit(:name, :sql, :description).merge(hidden: false))
+
+    group_ids = params.require(:query)[:group_ids]
+    DataExplorer::QueryGroup.where.not(group_id: group_ids).delete_all
+    group_ids&.each do |group_id|
+      query.query_groups.find_or_create_by!(group_id: group_id)
+    end
```
tgxworld, line 110:
```suggestion
    ActiveRecord::Base.transaction do
      query.update!(params.require(:query).permit(:name, :sql, :description).merge(hidden: false))

      group_ids = params.require(:query)[:group_ids]
      DataExplorer::QueryGroup.where.not(group_id: group_ids).delete_all
      group_ids&.each do |group_id|
        query.query_groups.find_or_create_by!(group_id: group_id)
      end
    end
```

app/controllers/data_explorer/query_controller.rb
```diff
@@ -0,0 +1,236 @@
...
+
+    render_serialized query, DataExplorer::QuerySerializer, root: 'query'
+  rescue DataExplorer::ValidationError => e
+    render_json_error e.message
+  end
+
+  def destroy
+    query = DataExplorer::Query.where(id: params[:id]).first_or_initialize
```
tgxworld, line 118: Hmm this doesn't make sense to me because this is a destroy action. If the `params[:id]` is not valid, we shouldn't have to create an instance of the object.

app/serializers/data_explorer/query_serializer.rb
```diff
@@ -0,0 +1,16 @@
+class DataExplorer::QuerySerializer < ActiveModel::Serializer
+  attributes :id, :sql, :name, :description, :param_info, :created_at, :username, :group_ids, :last_run_at, :hidden, :user_id
+
+  def param_info
+    object.params.map(&:to_hash) rescue nil
+  end
+
+  def username
+    object.user.username rescue nil
```
tgxworld, line 9: Not a fan of using a blanket `rescue nil` because it swallows all errors.

app/serializers/data_explorer/query_serializer.rb
```diff
@@ -0,0 +1,16 @@
+class DataExplorer::QuerySerializer < ActiveModel::Serializer
+  attributes :id, :sql, :name, :description, :param_info, :created_at, :username, :group_ids, :last_run_at, :hidden, :user_id
+
+  def param_info
+    object.params.map(&:to_hash) rescue nil
```
tgxworld, line 5: Not a fan of using a blanket rescue nil because it swallows all errors.

app/serializers/data_explorer/query_serializer.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  def username
+    object.user.username rescue nil
+  end
+
+  def group_ids
+    object.groups.map(&:id)
```
tgxworld, line 13: Hmm it doesn't seem like we're eager loading `groups` anywhere so this might cause an N+1?

db/migrate/20200810053843_create_data_explorer_queries.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+class CreateDataExplorerQueries < ActiveRecord::Migration[6.0]
+  def up
+    create_table :data_explorer_queries do |t|
```
tgxworld, line 5: Does any of the columns need to be `null: false`?

db/migrate/20200810053843_create_data_explorer_queries.rb
```diff
@@ -0,0 +1,64 @@
...
+      t.datetime :last_run_at
+      t.boolean :hidden, default: false
+      t.timestamps
+    end
+
+    create_table :data_explorer_query_groups do |t|
+      t.integer :query_id
+      t.integer :group_id
+      t.index :query_id
+      t.index :group_id
```
tgxworld, line 19: I think a uniqueness index might work better here?

db/migrate/20200810053843_create_data_explorer_queries.rb
```diff
@@ -0,0 +1,64 @@
...
+    DB.exec <<~SQL
+      INSERT INTO data_explorer_queries(id, name, description, sql, user_id, last_run_at, hidden, created_at, updated_at)
+      SELECT 
+        (value::json->>'id')::integer,
+        value::json->>'name',
+        value::json->>'description',
+        value::json->>'sql',
+        (value::json->>'created_by')::integer,
+        CASE WHEN (value::json->'last_run_at')::text = 'null' THEN 
+          now()
```
tgxworld, line 31: If the query has not been run, shouldn't the column be null?

plugin.rb
```diff
@@ -31,33 +31,33 @@ module ::DataExplorer
...
+      user_is_a_member_of_group?(group)
+    end
+  end
+
+  add_to_class(:guardian, :group_and_user_can_access_query?) do |group, query|
     return false if !current_user
     return true if current_user.admin?
-    return user_is_a_member_of_group?(group) &&
-           query.group_ids.include?(group.id.to_s)
+    return user_is_a_member_of_group?(group) && query.groups.where(id: group.id).present?
```
tgxworld, line 60:
```suggestion
    return user_is_a_member_of_group?(group) && query.groups.exists?(id: group.id)
```

`exists?` is slightly more efficient here since we don't have to create an ActiveRecord object.

tgxworld:
@lis2 I realized that most of my comments are existing code that has been moved around into new files. Feel free to ignore any comments that is too troublesome to fix or does not make sense.

lis2:
@tgxworld I will try to fix them all. Main goal was to make that code better not just move stuff to separate tabled. Thank you for your review! :)

app/controllers/data_explorer/query_controller.rb
lis2: Queries can be assigned to specific group members:
image

And then you have reports tab under specific group which allows you to evaluate queries:
image

app/controllers/data_explorer/query_controller.rb
lis2: yeah, I know it sounds weird but this is correct.

Besically what is happening here is that we have default queries in lib/queries.rb. However we need to give user an ability to remove them - hide them, similar to delete/hide custom queries saved in database.

So when user click "delete" button we create empty row in queries with hidden flag set to true and query id which is respected by index action

app/controllers/data_explorer/query_controller.rb
lis2: Actually I will just select :id and :name instead of having whole BasicGroupSerializer

db/migrate/20200810053843_create_data_explorer_queries.rb
lis2: Only hidden, I will add it - reason is that empty rows to hide default system queries mentioned ⬆️

CvX:
@lis2 There are rubocop issues: https://github.com/discourse/discourse/runs/1034247899#step:19:18 and the specs fail: https://github.com/discourse/discourse/runs/1034247830#step:28:22

PR #68, opened 2020-09-07
FIX: Store query groups in a temp table when fixing ids

romanrizzi:
Reintroduces the migration removed in cdfc5d4, and fixes it to work with `query_groups`. Since the code is fairly complex, I moved into a rake task so it can be tested and make sure it works.


danielwaterworth (changes requested):
This has become quite a monster, I'm glad to see some tests for it. The only thing I spotted was that the sequence number on the query groups table isn't getting updated afterwards.

romanrizzi:
Done. Thanks for reviewing @danielwaterworth! 😄

danielwaterworth (approved):

PR #69, opened 2020-09-23
FIX: the filtered array was not being used

jjaffeux:


jjaffeux:
@eviltrout AFAICT it comes from https://github.com/discourse/discourse-data-explorer/commit/d82109b00bd1c747c06c8f74bfcf15a425051aa7 I think the intent was to use this file array, so we either use it or just remove all of it and use fileList, ATM this is dead code.

eviltrout (approved):

PR #71, opened 2020-10-07
FIX: Correct the .discourse-compatibility syntax

CvX:
Fixes the core build.


udan11 (approved):

nbianca (approved):

SamSaffron:
oh ... my bad, thanks for fixing!

PR #76, opened 2020-10-14
FIX: fix ids rake task should be evaluated only once

lis2:
Without `task('data_explorer:fix_query_ids').clear` rake task is evaluated twice

### DIFF
--- a/lib/tasks/fix_query_ids.rake
+++ b/lib/tasks/fix_query_ids.rake
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 desc 'Fix query IDs to match the old ones used in the plugin store (q:id)'
+
+task('data_explorer:fix_query_ids').clear
 task 'data_explorer:fix_query_ids' => :environment do
   ActiveRecord::Base.transaction do
     # Only queries with unique title can be fixed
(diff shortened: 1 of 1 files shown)


tgxworld (approved):

tgxworld:
@lis2 Once you add `ON COMMIT DROP` for the temp tables this PR is good to go.

lib/tasks/fix_query_ids.rake
```diff
@@ -25,6 +25,9 @@ task 'data_explorer:fix_query_ids' => :environment do
       SQL
       additional_conflicts = additional_conflicts.map(&:id)
 
+      DB.exec "DROP TABLE IF EXISTS tmp_data_explorer_queries"
+      DB.exec "DROP TABLE IF EXISTS tmp_data_explorer_query_groups"
```
tgxworld, line 29: We need to clean this up if the tables have been left around in the database.

tgxworld (changes requested):

tgxworld (approved):
Let's unblock ourselves with this fix first but I think we should continue to investigate why the rake task has been loaded twice.

PR #78, opened 2020-10-16
FEATURE: Include more enums

tshenry:
This adds several new enums to help Data Explorer users quickly reference the meaning of the more obscure column values. It also moves `reviewables.status` to the main enum list since the reviewables table has been established for quite some time now. We could probably remove `queued_posts.state` as well, but it does act as an example of how to handle tables that aren't fully established or those that have been recently removed.


SamSaffron:
Nice change!

SamSaffron (approved):

PR #82, opened 2020-11-25
FEATURE: destroy old hidden queries

techAPJ:
This commit adds a scheduled job to hard delete queries that are hidden and

- were last run more than 7 days ago
- were updated more than 7 days ago

### DIFF
--- a/app/jobs/scheduled/delete_hidden_queries.rb
+++ b/app/jobs/scheduled/delete_hidden_queries.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module Jobs
+  class DeleteHiddenQueries < ::Jobs::Scheduled
+    every 7.days
+
+    def execute(args)
+      return unless SiteSetting.data_explorer_enabled
+
+      DataExplorer::Query.where("id > 0")
+        .where(hidden: true)
+        .where("(last_run_at IS NULL OR last_run_at < :days_ago) AND updated_at < :days_ago", days_ago: 7.days.ago)
+        .delete_all
+    end
+  end
+end
--- a/spec/jobs/scheduled/delete_hidden_queries_spec.rb
+++ b/spec/jobs/scheduled/delete_hidden_queries_spec.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe Jobs::DeleteHiddenQueries do
+  before do
+    Jobs.run_immediately!
+    SiteSetting.data_explorer_enabled = true
+  end
+
+  it "will correctly destroy old hidden queries" do
+    DataExplorer::Query.create!(id: 1, name: "A", description: "A description for A", sql: "SELECT 1 as value", hidden: false, last_run_at: 2.days.ago, updated_at: 2.days.ago)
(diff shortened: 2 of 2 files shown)


app/jobs/scheduled/delete_hidden_queries.rb
```diff
@@ -0,0 +1,17 @@
...
+
+module Jobs
+  class DeleteHiddenQueries < ::Jobs::Scheduled
+    every 7.days
+
+    def execute(args)
+      return unless SiteSetting.data_explorer_enabled
+
+      DataExplorer::Query.where(hidden: true).each do |query|
+        if query.last_run_at < 7.days.ago && query.updated_at < 7.days.ago
```
davidtaylorhq, line 11: Can `last_run_at` be `nil`?

app/jobs/scheduled/delete_hidden_queries.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Jobs
+  class DeleteHiddenQueries < ::Jobs::Scheduled
+    every 7.days
+
+    def execute(args)
+      return unless SiteSetting.data_explorer_enabled
+
+      DataExplorer::Query.where(hidden: true).each do |query|
```
davidtaylorhq, line 10: Some sites have stacked up 10,000+ hidden queries, so we should probably do this in batches... Or alternatively do the `last_run_at` and `updated_at` checks in SQL, and use `.delete_all`

davidtaylorhq (commented):
Looks good. What happens if you press delete on one of the pre-seeded queries in the UI? Does [the deletion](https://github.com/discourse/discourse-data-explorer/blob/master/app/controllers/data_explorer/query_controller.rb#L102) persist? Or does the query re-appear after 7 days? I think we might need to skip destroying rows with ID < 0 🤔

techAPJ:
> Looks good. What happens if you press delete on one of the pre-seeded queries in the UI? Does [the deletion](https://github.com/discourse/discourse-data-explorer/blob/master/app/controllers/data_explorer/query_controller.rb#L102) persist? Or does the query re-appear after 7 days? I think we might need to skip destroying rows with ID < 0 🤔

The pre-seeded queries can not be deleted from UI because we hide the delete button for them.

app/jobs/scheduled/delete_hidden_queries.rb
techAPJ: Good idea. Will make it so.

app/jobs/scheduled/delete_hidden_queries.rb
techAPJ: Yes, let me check for the presence of `last_run_at`.

davidtaylorhq:
> The pre-seeded queries can not be deleted from UI because we hide the delete button for them.

Hmm... I think it is supposed to work. The reason the 'hidden' feature was added was [to delete system queries](https://meta.discourse.org/t/how-to-delete-system-queries-in-data-explorer/147524/8?u=david). On localhost I see the delete button for default queries. (and I can also edit them... which seems unintended... but that's a different problem)

techAPJ:
@davidtaylorhq Updated the PR per your feedback.

davidtaylorhq (approved):

PR #84, opened 2020-12-07
FIX: Use standard error handling for requests, special-case moderators

riking:
By removing the catch and letting the error propagate to the top-level error handler, we can receive more detailed error reports including exactly which request failed.

The assumption that any request failure is a permissions error is replaced by an explicit client-side check before making any requests.


riking:
Meta topic https://meta.discourse.org/t/the-data-explorer-is-only-available-to-admins/94801/

featheredtoast (approved):
linting aside, this looks good :+1:

PR #85, opened 2020-12-22
DEV: Use the new discourseDebounce function wrapper.

romanrizzi:
We recently merged a Discourse core's PR to replace usages of Ember's debounce and discourseDebounce with a new debounce wrapper. The new wrapper works exactly like Ember's debounce but internally calls "run" when called in test mode.

This PR replaces all usages of other debounce functions with the new wrapper and fallbacks to Ember's debounce for backward-compatibility.

### DIFF
--- a/assets/javascripts/discourse/components/explorer-schema.js.es6
+++ b/assets/javascripts/discourse/components/explorer-schema.js.es6
@@ -2,7 +2,8 @@ import {
   default as computed,
   observes,
 } from "discourse-common/utils/decorators";
-import debounce from "discourse/lib/debounce";
+import discourseDebounce from "discourse-common/lib/debounce";
+import debounce from "@ember/runloop";
 
 export default Ember.Component.extend({
   actions: {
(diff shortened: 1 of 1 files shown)


eviltrout (approved):

goodeas:
Hi @romanrizzi @eviltrout 

Just wondering if this change was tested on Discourse stable? I am running a self-hosted instance, now v2.6.2 (eb9f15a94a), also with the latest Data Explorer updates as of today. I am unable to edit any queries in Data Explorer - the JS error seems to be related to this change. I described the issue and the error a couple of weeks ago in the following post:

https://meta.discourse.org/t/data-explorer-plugin/32566/214

I'm still seeing the same symptoms after the latest updates today.

Thanks
Andrew

romanrizzi:
Hi @goodeas

Sorry for the trouble. Should be fixed in 50e62d84e.

goodeas:
Thanks @romanrizzi

PR #92, opened 2021-02-11
FIX: Show badge names in query results

AndrewPrigorshnev:


assets/javascripts/discourse/templates/explorer/badge.raw.hbs
```diff
@@ -3,4 +3,4 @@
    class="user-badge {{badge.badgeTypeClassName}}"
    title="{{badge.displayDescription}}"
    data-badge-name="{{badge.name}}">{{icon-or-image badge.icon}}
-<span class="badge-display-name">{{badge.displayName}}</span></a>
```
ZogStriP, line 6: I think instead of using `name` which might not always be user friendly, it would be better to add the [display_name](https://github.com/discourse/discourse/blob/956f8492507f6004272b1e6900f3e56c5c0f41a1/app/models/badge.rb#L256-L258) property to the [`small_badge_serializer`](https://github.com/discourse/discourse-data-explorer/blob/master/app/serializers/data_explorer/small_badge_serializer.rb).

ZogStriP (changes requested):
Love that you added a test 👍 But I think this feature should be fixed server-side 😉 

You can definitely keep the front-end test though.

assets/javascripts/discourse/templates/explorer/badge.raw.hbs
AndrewPrigorshnev: You're right, I've fixed this.

app/serializers/data_explorer/small_badge_serializer.rb
```diff
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
 class DataExplorer::SmallBadgeSerializer < ApplicationSerializer
-  attributes :id, :name, :badge_type, :description, :icon
+  attributes :id, :name, :display_name, :badge_type, :description, :icon
+
+  def display_name
+    object&.name
```
ZogStriP, line 7: Isn't `display_name` available? What's the type of `object`?

AndrewPrigorshnev: OMG 🤦. I had a false feeling that it could be a super-fast fix. `display_name` is available indeed. I've fixed it and checked it with a changed locale.

assets/javascripts/discourse/templates/explorer/badge.raw.hbs
```diff
@@ -1,6 +1,6 @@
 {{! source: badge-button component }}
 <a href="{{baseuri}}/badges/{{badge.id}}/{{badge.name}}"
    class="user-badge {{badge.badgeTypeClassName}}"
-   title="{{badge.displayDescription}}"
+   title="{{badge.display_name}}"
```
AndrewPrigorshnev, line 4: one more fixed property

ZogStriP (approved):
LGTM 👍

PR #93, opened 2021-02-11
FEATURE: Add a graph report to query results

AndrewPrigorshnev:


assets/javascripts/discourse/components/bar-chart.js.es6
```diff
@@ -0,0 +1,96 @@
...
+  didInsertElement() {
+    this._super(...arguments);
+    this._initChart();
+  },
+
+  didUpdate() {
+    this._super(...arguments);
+    this.chart.data = this.get("data");
+    this.chart.update();
+  },
```
AndrewPrigorshnev, line 81: Right now, this hook never calls because the query-result component always reinserts after querying data. But it's better to have it here to escape regressions in the future.

test/javascripts/unit/components/query-result-test.js.es6
```diff
@@ -0,0 +1,103 @@
+import { moduleFor } from "ember-qunit";
+import { test } from "qunit";
+
+moduleFor("component:query-result");
+
```
AndrewPrigorshnev, line 5: I see that there are just a few unit tests for components in Discourse. But we need to test data for the chart somehow. Chart renders on canvas, so we can't check the rendered result. The solution is to check computed properties. As I know, it's impossible to do it in integration tests. That's why I'm using unit tests here.

assets/javascripts/discourse/templates/explorer-query-result.hbs
```diff
@@ -18,20 +25,27 @@
...
-  <thead>
-  <tr class="headers">
-    {{#each columnDispNames as |col|}}
-      <th>{{col}}</th>
+{{#if showGraph}}
+  {{bar-chart
+    labels=graphLabels
+    values=graphValues
+    datasetName=graphDatasetName}}
+{{else}}
```
AndrewPrigorshnev, line 33: Speaking of naming. We use the word "Graph" in UI and chart.js library for drawing charts. It's unclear where to stop using the word "Graph" and start using the word "Chart." Maybe it's better just to use the word "Chart" everywhere?

assets/javascripts/discourse/components/bar-chart.js.es6
ZogStriP: I'd rather have a test than some code that is unused but might be used. Much better at preventing regressions as well 😉

ZogStriP:
@jjaffeux & @CvX would you mind having a look at this PR?

assets/javascripts/discourse/components/bar-chart.js.es6
```diff
@@ -0,0 +1,97 @@
...
+
+  didInsertElement() {
+    this._super(...arguments);
+    this._initChart();
+  },
+
+  didUpdate() {
+    this._super(...arguments);
+    this.chart.data = this.get("data");
```
jjaffeux, line 79: do we need get here ?

assets/javascripts/discourse/components/bar-chart.js.es6
```diff
@@ -0,0 +1,97 @@
...
+  willDestroyElement() {
+    this._super(...arguments);
+    this.chart.destroy();
+  },
+
+  _initChart() {
+    loadScript("/javascripts/Chart.min.js").then(() => {
+      const canvas = this.element.querySelector("canvas");
+      const context = canvas.getContext("2d");
+      const config = this.get("config");
```
jjaffeux, line 92: same here, do we need get ?

assets/javascripts/discourse/components/query-result.js.es6
```diff
@@ -139,6 +140,61 @@ const QueryResultComponent = Ember.Component.extend({
...
+    "rows.[]",
+    "content.colrender.[]",
+    "content.result_count",
+    "colCount"
+  )
+  canShowGraph(rows, colrender, result_count, colCount) {
+    const hasTwoColumns = colCount === 2;
+    const secondColumnContainsNumber =
+      result_count > 0 && typeof rows[0][1] === "number";
+    const secondColumnContainsId = colrender[1] !== undefined;
```
jjaffeux, line 153: de we explicitly need to check for undefined here ?

assets/javascripts/discourse/components/query-result.js.es6
```diff
@@ -139,6 +140,61 @@ const QueryResultComponent = Ember.Component.extend({
...
+
+    return (
+      hasTwoColumns && secondColumnContainsNumber && !secondColumnContainsId
+    );
+  },
+
+  @computed("content.rows.[]", "content.colrender.[]")
+  graphLabels(rows, colRender) {
+    const labelSelectors = new Map([
```
jjaffeux, line 162: could you explain the need for a Map here ?

assets/javascripts/discourse/components/bar-chart.js.es6
```diff
@@ -0,0 +1,97 @@
...
+  barsHoverColor: themeColor("--tertiary-high"),
+  gridColor: themeColor("--primary-low"),
+  labelsColor: themeColor("--primary-medium"),
+  chart: undefined,
+
+  @computed("data", "options")
+  config(data, options) {
+    return {
+      type: "bar",
+      data: data,
```
jjaffeux, line 16:
```suggestion
      data,
```

assets/javascripts/discourse/components/bar-chart.js.es6
```diff
@@ -0,0 +1,97 @@
...
+      options: options,
```
jjaffeux, line 17:
```suggestion
      options,
```

assets/javascripts/discourse/components/bar-chart.js.es6
```diff
@@ -0,0 +1,97 @@
...
+    };
+  },
+
+  @computed("labels.[]", "values.[]", "datasetName")
+  data(labels, values, datasetName) {
+    return {
+      labels: labels,
```
jjaffeux, line 24:
```suggestion
      labels,
```

assets/javascripts/discourse/components/query-result.js.es6
```diff
@@ -139,6 +140,61 @@ const QueryResultComponent = Ember.Component.extend({
...
+
+    return rows.map((r) => this._cutGraphLabel(r[0]));
+  },
+
+  @computed("content.rows.[]")
+  graphValues(rows) {
+    return rows.map((r) => r[1]);
+  },
+
+  @computed("columnDispNames.[]")
```
jjaffeux, line 193: same minor nitpick here, but you can probably do:

```
graphDatasetName: reads("columnDispNames.1")
```

assets/javascripts/discourse/components/query-result.js.es6
```diff
@@ -211,13 +267,23 @@ const QueryResultComponent = Ember.Component.extend({
     });
   },
 
+  _cutGraphLabel(label) {
```
jjaffeux, line 270: minor, I think I would go with ellipsis in the name instead of cut

assets/javascripts/discourse/components/bar-chart.js.es6
```diff
@@ -0,0 +1,97 @@
+import loadScript from "discourse/lib/load-script";
+import { default as computed } from "discourse-common/utils/decorators";
+import themeColor from "../lib/themeColor";
+
+export default Ember.Component.extend({
+  barsColor: themeColor("--tertiary"),
+  barsHoverColor: themeColor("--tertiary-high"),
+  gridColor: themeColor("--primary-low"),
+  labelsColor: themeColor("--primary-medium"),
+  chart: undefined,
```
jjaffeux, line 10: I would prefer null here, unless you have a specific reason to use undefined ?

test/javascripts/unit/components/query-result-test.js.es6
```diff
@@ -0,0 +1,103 @@
...
+      ["user1", 10],
+      ["user2", 20],
+    ],
+  };
+  this.subject().setProperties({
+    content: results,
+  });
+
+  assert.deepEqual(
+    this.subject().get("graphLabels"),
```
jjaffeux, line 21: Won't note here everywhere but get is probably not needed each time you use it, if subject is null get won't save you

assets/javascripts/discourse/components/bar-chart.js.es6
AndrewPrigorshnev: You're right, we don't need `get` here and in all other places.

assets/javascripts/discourse/components/query-result.js.es6
```diff
@@ -139,6 +140,61 @@ const QueryResultComponent = Ember.Component.extend({
...
+          const label = labelSelector(relation);
+          return this._cutGraphLabel(label);
+        });
+      }
+    }
+
+    return rows.map((r) => this._cutGraphLabel(r[0]));
+  },
+
+  @computed("content.rows.[]")
```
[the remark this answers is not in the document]
AndrewPrigorshnev: It's a good suggestion. Thanks.

assets/javascripts/discourse/components/query-result.js.es6
AndrewPrigorshnev: This one is good too.

assets/javascripts/discourse/components/query-result.js.es6
AndrewPrigorshnev: Good idea.

assets/javascripts/discourse/components/bar-chart.js.es6
AndrewPrigorshnev: Definitely `null` is better for an absent object.

assets/javascripts/discourse/components/query-result.js.es6
AndrewPrigorshnev: I just needed a hash-like structure to save these selectors in an easy-to-read way and to access them. I've chosen `Map` arbitrarily. A plain object will work perfectly well here - the keys are strings, and we don't use any special methods of `Map`.  

Well, I think it'll be simpler with a plain object, and I would like to use it. Is this what concerns you here?

And I consciously allocated it on the function level, not on the object level, in favor of readability. Is this alright?

assets/javascripts/discourse/components/query-result.js.es6
AndrewPrigorshnev: It would work without an explicit check.

But I feel that with an explicit check it's easier to understand what's going on here. But I can change this if it is more idiomatic for the Discourse codebase:
```
const secondColumnContainsId = colrender[1];

return (
  hasTwoColumns && secondColumnContainsNumber && !secondColumnContainsId
);
```

assets/javascripts/discourse/components/bar-chart.js.es6
AndrewPrigorshnev: I agree that it's better to have a test. We could check that the chart is updating when we're performing the query one more time. Unfortunately, the chart is rendered on a canvas, and it's hard to check if it's updated. It's possible to [check pixels on a canvas](https://www.npmjs.com/package/qunit-assert-canvas), but it looks overcomplicated to me.

Probably we can remove this hook. But then the problem would appear if we stop reinserting the whole component with query results after every query or if someone uses this code or the component itself for a chart that should be updated without reinserting.

assets/javascripts/discourse/components/query-result.js.es6
```diff
@@ -187,12 +188,7 @@ const QueryResultComponent = Ember.Component.extend({
 
   @computed("content.rows.[]")
   graphValues(rows) {
-    return rows.map((r) => r[1]);
-  },
-
-  @computed("columnDispNames.[]")
-  graphDatasetName(columnDispNames) {
-    return columnDispNames[1];
+    return rows.mapBy(1);
```
jjaffeux, line 191: there's a mapBy computed property in ember https://api.emberjs.com/ember/release/functions/@ember%2Fobject%2Fcomputed/mapBy

assets/javascripts/discourse/components/query-result.js.es6
jjaffeux: Yes I think a js object would be enough, we don't use Map much unless the order matters

assets/javascripts/discourse/components/query-result.js.es6
```diff
@@ -139,6 +142,51 @@ const QueryResultComponent = Ember.Component.extend({
     return transformedRelTable(groups);
   },
 
+  @computed(
+    "rows.[]",
+    "content.colrender.[]",
+    "content.result_count",
+    "colCount"
+  )
+  canShowGraph(rows, colrender, result_count, colCount) {
```
ZogStriP, line 151: We use camelCase in JS as much as possible

```suggestion
  canShowGraph(rows, colRender, resultCount, colCount) {
```

assets/javascripts/discourse/templates/explorer-query-result.hbs
ZogStriP: Yes, we should be consistent and use "chart" everywhere in the code but keep "graph" in the UI since that's much "user-friendly".

jjaffeux:
coupd you give bar-chart component a different bame to avoid any name clash with core in the future please? something like data-explorer-bar-chart

assets/javascripts/discourse/templates/explorer-query-result.hbs
AndrewPrigorshnev: Fine, I've renamed this in code and in the test messages.

test/javascripts/acceptance/run-query-test.js.es6
```diff
@@ -0,0 +1,193 @@
...
+        ],
+      });
+    });
+  });
+
+  test("it runs query and renders data and a chart", async function (assert) {
+    await visit("admin/plugins/explorer?id=-6");
+
+    assert.ok(
+      queryAll("div.name h1").text().trim() === "Top 100 Likers",
```
ZogStriP, line 162: Do we need `queryAll` when we're looking for only 1 element?

AndrewPrigorshnev: Not really 😊. In most cases here, `query()` is enough. I've fixed this.

ZogStriP (approved):
LGTM 👍 

I'll merge it tomorrow.

AndrewPrigorshnev:
@ZogStriP this evening a problem with CI actions appears. They fail on the step "Migrate database" with a message "libffi.so.6: cannot open shared object file: No such file or directory". 

Runners started using Ubuntu 20.04.2 and [here](https://stackoverflow.com/a/66099514/3206146) is explained what's going on. Please let me know if I can help with this.

AndrewPrigorshnev:
@ZogStriP @jjaffeux I've discovered just today that there are special buttons for re-requesting review, to let you know that I've applied your suggestion. I haven't used GitHub a lot for pull-requests before.

Sorry if you were waiting for my replies.

ZogStriP:
@AndrewPrigorshnev no worries at all. GitHub will send us notifications whenever you comment. So we've be pretty much up to date 😉

ZogStriP:
@AndrewPrigorshnev I merged another PR before this and now there's a conflict. Can you fix it?

AndrewPrigorshnev:
Sure, will do this soon.

PR #101, opened 2021-03-02
Allow admins to see the group reports nav item

graydenshand:
Previously, the report tab would appear only if the user was a member of the group. With this change, admins can see the reports in any group.


graydenshand:
Bueller..? Please accept this pull request or let me know what I need to change for it to be accepted.

davidtaylorhq (changes requested):
Sorry for the delay here @graydenshand - overall the fix looks good, but there is a 'prettier' failure showing in the CI checks. Can you try running prettier against your changes, and commit the result?

davidtaylorhq (approved):
Looks great, thanks!

graydenshand:
Thanks for your help with this @davidtaylorhq!

PR #107, opened 2021-04-08
DEV: Improve logic for showing/hiding the reports tab in group screens

davidtaylorhq:
Previously this was adding an extra AJAX request to check if the group had any queries available. Now a boolean is included in the group serializer, so there is no need for the extra request.

Removing this ajax request will also stop other plugin JS integration tests from failing when the data-explorer plugin is installed.

This commit also fixes the HTML markup of the tab, so that it doesn't have a `<ul>` nested inside the existing `<ul>`. Also adds an icon for good measure.


plugin.rb
```diff
@@ -60,6 +60,14 @@ def self.plugin_name
     return user_is_a_member_of_group?(group) && query.groups.exists?(id: group.id)
   end
 
+  add_to_serializer(:group_show, :has_visible_data_explorer_queries, false) do
+    DataExplorer::Query.for_group(object).exists?
+  end
+
+  add_to_serializer(:group_show, :include_has_visible_data_explorer_queries?, false) do
+    SiteSetting.data_explorer_enabled && scope.user_is_a_member_of_group?(object)
```
pmusaraj, line 68: Should there be a spec for this?

pmusaraj (approved):

plugin.rb
davidtaylorhq: Yeah there probably should - good call. Added some specs

PR #56, opened 2020-07-23
UX: Add additional instructions in modal help menu on query editor page 

RickyC0626:

PR #109, opened 2021-04-12
UX: Add a help modal dialog

AndrewPrigorshnev:
This PR is based on [this closed PR](https://github.com/discourse/discourse-data-explorer/pull/56).

### DIFF
--- a/assets/javascripts/discourse/templates/modal/query-help.hbs
+++ b/assets/javascripts/discourse/templates/modal/query-help.hbs
@@ -2,7 +2,50 @@
   <h3>{{i18n "explorer.help.modal.custom_params.title"}}</h3>
   <p>{{i18n "explorer.help.modal.custom_params.body"}}</p>
   <pre><code class="sql">{{html-safe (i18n "explorer.help.modal.custom_params.code")}}</code></pre>
+  <strong><p>{{i18n "explorer.help.modal.custom_params.body2"}}</p></strong>
+
+  <p>{{i18n "explorer.help.modal.custom_params.body3"}}</p>
+  <pre><code class="sql">{{html-safe (i18n "explorer.help.modal.custom_params.code2")}}</code></pre>
+  <hr>
 
   <h3>{{i18n "explorer.help.modal.query_data_types.title"}}</h3>
   <p>{{i18n "explorer.help.modal.query_data_types.body"}}</p>
+  <table class="topic-list">
+    <thead>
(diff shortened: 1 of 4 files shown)


assets/javascripts/discourse/templates/modal/query-help.hbs
```diff
@@ -0,0 +1,50 @@
+{{#d-modal-body title="explorer.help.modal_title"}}
+  <h2>{{i18n "explorer.help.auto_resolution.title"}}</h2>
```
ZogStriP, line 2: I wonder if the modal content should just be one translatable string instead?

AndrewPrigorshnev: Yeah, it'll definitely make things simpler.

AndrewPrigorshnev: @ZogStriP  I wonder if it's all right to put that much HTML in a translatable string when putting all modal content into it. The modal even contains a table:

Screenshot 2021-04-13 at 15 06 03

AndrewPrigorshnev: I've split the modal content into 4 pieces and eventually removed the table and started using a list instead. It's a bit tricky business to use a table here, we need a CSS class to be applied to the table, so we can't put the whole table into a resource. With the list it looks like this:

Screenshot 2021-04-15 at 17 21 47

config/locales/client.en.yml
```diff
@@ -14,11 +14,39 @@ en:
...
+        custom_params: "<h2>Creating Custom Parameters</h2>
+          <p>To create custom parameters for your queries, put this at the top of your query and follow the format:</p>
+          <pre><code>-- [params]\n-- int :num = 1\n\nSELECT :num</code></pre>
+          <p><i>Note: the first line with [params] is required, along with two dashes preceding it and every
+          custom parameter you want to declare.</i></p>"
+        default_values: "<h3>Default Values</h3>
+          <p>You can declare parameters with or without default values. Default values will show up in a text field
+          below the query editor, which you can edit to your needs. Parameters declared without default values will
+          still generate a text field, but will be empty and highlighted red.</p>
+          <pre><code>-- [params]\n-- text :username = my_username\n-- int :age</code></pre>"
```
riking, line 35: Additionally, several parameter types accept easily paste-able text formats or provide lookup assistance.

Here is the full list of parameter types, which we _do_ need to document somewhere: https://github.com/discourse/discourse-data-explorer/blob/main/plugin.rb#L719-L830

PR #126, opened 2021-07-07
FEATURE: open a system file picker directly instead of opening the import modal

AndrewPrigorshnev:
To import a query into Data Explorer you need to push the button and then deal with the import modal:

Screenshot 2021-07-01 at 14 49 59

Screenshot 2021-07-01 at 14 50 05

Instead, we want just to be triggering a system file picker directly. This PR makes it happen.

In the beginning, I wanted just to move code around to fix this problem fastly but unfortunately, we have a big [controller](https://github.com/discourse/discourse-data-explorer/blob/main/assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6) + template that makes it hard to do it without introducing technical debt. Eventually, I added a small component `upload-files-button`. This is a button that fires a system file picker and has an event `onFilesChosen`, the enclosing component (or controller) can subscribe to it. I think it's a low-pri thing but probably it's worth moving this component to core later to use it everywhere where we need a button that fires a system file-picker.

Sadly, tests are skipped. I was able to trigger an uploading event in tests, but for some reason when running tests the component receives an empty list of files. I couldn't make them work and I'm going to revisit them later.


ZogStriP:
> I think it's a low-pri thing but probably it's worth moving this component to core later to use it everywhere where we need a button that fires a system file-picker.

I think we should do that first and then use it here in this plugin.

AndrewPrigorshnev:
> I think we should do that first and then use it here in this plugin

Cool, I'll do it then. 

I'm closing this PR so nobody will spend time reviewing it. I'll open a new one here later when I merge a component into core.

PR #127, opened 2021-07-16
FEATURE: get rid of the import a query modal

AndrewPrigorshnev:
This is the new version of https://github.com/discourse/discourse-data-explorer/pull/126. Now the `pick-files-button` moved to core. 

To import a query into Data Explorer, you need to push the button and then deal with the import modal:

124735139-d168f500-df26-11eb-8b71-82f0ada18129

124735161-d75ed600-df26-11eb-943b-024c873df6da

Instead, we want just to be triggering a system file picker directly. This PR makes it happen.


assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
```diff
@@ -30,6 +30,16 @@ export default Ember.Controller.extend({
   sortBy: ["last_run_at:desc"],
   sortedQueries: Ember.computed.sort("model", "sortBy"),
 
+  @computed
+  acceptedImportFileTypes() {
+    return [
```
martin-brennan, line 35: I am just wondering why we need all these variants of the JSON type? Do different browsers send different things? Usually it is application/json

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
```diff
@@ -117,16 +127,41 @@ export default Ember.Controller.extend({
...
+    const reader = new FileReader();
+    reader.onload = (evt) => {
+      this._importQuery(evt.target.result);
+    };
+
+    reader.readAsText(file);
+  },
+
+  _importQuery(file) {
+    const query = JSON.parse(file).query;
```
martin-brennan, line 140: I think we should add some JSON parse error handling here, in case the JSON in the file is not correctly structured. Then you can show a message to the user and reset the `loading` state

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
```diff
@@ -117,16 +127,41 @@ export default Ember.Controller.extend({
       .finally(() => this.set("loading", false));
   },
 
+  _readQueryFile(file) {
+    const reader = new FileReader();
+    reader.onload = (evt) => {
```
martin-brennan, line 132: Probably want to implement `FileReader.onerror` here as well to be able to reset the state and show a message on error. If this is a pattern we use in a few places in future we may want to refactor this out into a `lib` that we can import. No need for that yet though as this is the only place we are doing this so far :)

martin-brennan: E.g. `app/lib/file-reader.js`

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
```diff
@@ -117,16 +127,41 @@ export default Ember.Controller.extend({
...
+
+    this.set("loading", true);
+    this.store
+      .createRecord("query", query)
+      .save()
+      .then((q) => {
+        this.set("loading", false);
+        this.addCreatedRecord(q.target);
+      })
+      .catch(popupAjaxError);
```
martin-brennan, line 151: You will need to set `loading` to false here too I assume?

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
martin-brennan: Also do we know what the structure should look like ahead of time? Maybe we should check for the keys that are required and throw an error if the user doesn't provide them?

martin-brennan (changes requested):
Looking good and I like the new component, just a few questions.

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
AndrewPrigorshnev: I don't know initial intentions behind this list, I've just moved it from other place in this PR. But looks like you're right and it's more than enough to have just `application/json`. For example, [text/javascript is obsolete, and application/x-javascript was experimental ](https://stackoverflow.com/a/9664327/3206146).

AndrewPrigorshnev:
All comments about error handling and setting `loading` to false are slightly out of the scope of this PR, I wanted just to get rid of an unnecessary modal and carefully move existing code. 

At the same time, they all are perfect catches, and we definitely need to fix them. So I fixed these problems and I also switched these methods to `async/await` pattern, all these checks implemented with promise chaining would look overcomplicated.

Speaking of extracting code for reading files to `lib`, this wrapper that can be awaited can be our starting point:
```javascript
  _readFileAsTextAsync(file) {
    return new Promise((resolve, reject) => {
      const reader = new FileReader();
      reader.onload = () => {
        resolve(reader.result);
      };
      reader.onerror = reject;

      reader.readAsText(file);
    });
  }
```

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
AndrewPrigorshnev: And [the mime type for json text is application/json](https://stackoverflow.com/a/477819/3206146).

We also already use just `application/json` with file inputs in several places, it seems to work fine.

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
```diff
@@ -117,16 +124,64 @@ export default Ember.Controller.extend({
...
+    const parsed = JSON.parse(json);
+    const query = parsed.query;
+    if (!query || !query.sql) {
+      throw new TypeError();
+    }
+    query.id = 0; // 0 means no Id yet
+    return query;
+  },
+
+  _readFileAsTextAsync(file) {
```
martin-brennan, line 145: Shouldn't this function be marked `async` if you are `await`ing it in `_importQuery`?

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
```diff
@@ -117,16 +124,64 @@ export default Ember.Controller.extend({
...
 
+  async _importQuery(file) {
+    const json = await this._readFileAsTextAsync(file);
+    const query = await this._parseQuery(json);
+    const record = this.store.createRecord("query", query);
+    const response = await record.save();
+    return response.target;
+  },
+
+  async _parseQuery(json) {
```
martin-brennan, line 135: I don't think this function needs to be `async`, nothing that is promise-related is happening here.

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
```diff
@@ -117,16 +124,64 @@ export default Ember.Controller.extend({
...
 
-    importQuery() {
-      showModal("import-query");
-      this.set("showCreate", false);
+    import(files) {
+      this.set("loading", true);
+      const file = files[0];
+      this._importQuery(file)
+        .then((record) => this.addCreatedRecord(record))
+        .catch((e) => {
```
martin-brennan, line 169: Much better, thanks for adding the error handling

config/locales/client.en.yml
```diff
@@ -15,6 +15,8 @@ en:
       import:
         label: "Import"
         modal: "Import A Query"
+        unparseable_json: "Unparseable json file."
+        wrong_json: "Wrong json file. A json file should contain an 'query' object, which should at least has a 'sql' property."
```
martin-brennan, line 19: Nitpick, should be `"which should at least have an 'sql' property"`. Also...I don't mind if you change this or not, it's just another nitpick, but I think JSON should be capitalised because it is an acronym

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
```diff
@@ -30,6 +32,11 @@ export default Ember.Controller.extend({
   sortBy: ["last_run_at:desc"],
   sortedQueries: Ember.computed.sort("model", "sortBy"),
 
+  @computed
+  acceptedImportFileTypes() {
+    return ["application/json"];
```
martin-brennan, line 37: Thanks for changing this. I think the other modal was done in 2014 so maybe this wasn't the accepted mime type back then? Anyway application/json is totally fine, never used anything but that.

martin-brennan (changes requested):
Much better, still a few minor changes though.

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
AndrewPrigorshnev: Oh, this is redundant here, a good catch!

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
AndrewPrigorshnev: No, we don't need it here. Normally, you need `async` if you're awaiting something _inside_ the function. This function just create and return a promise, and we're awaiting that promise outside.

config/locales/client.en.yml
AndrewPrigorshnev: I've changed both, thank you 👍, please feel free to comment such mistakes.

assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
martin-brennan: Ah cool I didn't realise that thanks for clarifying

martin-brennan (approved):
LGTM :+1:

PR #91, opened 2021-02-09
DEV: Update CI workflows

discoursebot:
Updates CI from discourse/.github

PR #129, opened 2021-07-30
DEV: make tests work in Ember CLI

AndrewPrigorshnev:


test/javascripts/unit/components/query-result-test.js.es6
```diff
@@ -1,99 +1,94 @@
-import { moduleFor } from "ember-qunit";
-import { test } from "qunit";
+import {module, test} from "qunit";
+import {setupTest} from "ember-qunit";
```
CvX, line 2: TIL: we need to configure the plugin linting action to run prettier on `/test` directory 😉

AndrewPrigorshnev: Yep, I think I'll take a look now.

AndrewPrigorshnev: Hmm, I'm pretty sure this was working before. Probably this https://github.com/discourse/discourse-data-explorer/pull/91 broke it but at first glance it runs prettier on `/test` directory 🤔.

CvX: Looks like `**` in `ls test/**/*.@(js|es6)` doesn't traverse multiple levels of directories…
So it would have to be `ls test/javascripts/*/*.@(js|es6)` unless that somehow doesn't work for other plugins.

CvX: Fixed in https://github.com/discourse/.github/pull/39 🙂

PR #130, opened 2021-08-10
FIX: back button

AndrewPrigorshnev:
Back button wasn't working when running in Ember CLI:

Screenshot 2021-08-10 at 20 04 31

This PR fixes it by doing the transition to the same route but without URL parameter in a [proper way](https://guides.emberjs.com/v3.12.0/routing/query-params/#toc_transitionto). This works in both our environments (legacy-3.12.0, and ember-cli-3.15.0).

I wanted to go further and start using a [full transition](https://guides.emberjs.com/v3.12.0/routing/query-params/#toc_opting-into-a-full-transition) by adding this to the route:
```javascript
queryParams: { 
    filter: { 
        refreshModel: true
    }
},
```
This way `goHome` method could be much simpler:
``` javascript
    goHome() {
      /* we'd remove setting properties then
      this.setProperties({
        asc: null,
        order: null,
        showResults: false,
        editDisabled: false,
        selectedQueryId: null,
        sortBy: ["last_run_at:desc"],
      });
     */ 
      this.transitionToRoute({ queryParams: { id: null } });
    },
```

Unfortunately, this controller is quite fragile and this breaks another transition. I feel like it's better to leave it as is by now, but I'm going to address this some day later.

### DIFF
--- a/assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
+++ b/assets/javascripts/discourse/controllers/admin-plugins-explorer.js.es6
@@ -155,8 +155,7 @@ export default Ember.Controller.extend({
         selectedQueryId: null,
         sortBy: ["last_run_at:desc"],
       });
-      this.send("refreshModel");
-      this.transitionToRoute("adminPlugins.explorer");
+      this.transitionToRoute({ queryParams: { id: null } });
     },
 
     showHelpModal() {
--- a/assets/javascripts/discourse/routes/admin-plugins-explorer.js.es6
+++ b/assets/javascripts/discourse/routes/admin-plugins-explorer.js.es6
@@ -40,11 +40,4 @@ export default DiscourseRoute.extend({
   setupController(controller, model) {
     controller.setProperties(model);
   },
-
-  actions: {
-    refreshModel() {
-      this.refresh();
-      return false;
-    },
-  },
 });


jjaffeux:
What was the error please?

AndrewPrigorshnev:
> What was the error please?

There wasn't an error in the console. 

Here is what was happening as I understand this. When we push this button we go from the route `plugins/explorer?id=1` to `plugins/explorer` (without query param). It is said in docs for Ember 3.12 that query param values [are sticky](https://guides.emberjs.com/v3.12.0/routing/query-params/#toc_sticky-query-param-values):

> if you make changes to a query param and then leave and re-enter the route, the new value of that query param will be preserved (rather than reset to its default)

We were doing the transition without specifying query params:
```javascript
this.transitionToRoute("adminPlugins.explorer");
```
According to the docs that should have lead to using the previous query param `?id=1` and transition shouldn't have been hapenning. But for some reason that wasn't the case in the 3.12 version. But the transition doesn't happen anymore in Ember 3.15, probably some bug was fixed.

It's reproducible on Meta now:
https://user-images.githubusercontent.com/1274517/128926870-e523ef3a-3a61-47a5-bb7e-eb40e400178f.mov

jjaffeux (approved):

PR #153, opened 2022-01-19
FEATURE: Move query limit to hidden site setting

martin-brennan:
Previously the `QUERY_RESULT_DEFAULT_LIMIT` const was used
to limit the number of query results. This commit adds the
`data_explorer_query_result_limit` site setting which defaults
to 1000 and has a max of 10000 which matches the const
`QUERY_RESULT_MAX_LIMIT`.

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -25,7 +25,8 @@ end
 add_admin_route 'explorer.title', 'explorer'
 
 module ::DataExplorer
-  QUERY_RESULT_DEFAULT_LIMIT = 1000
+  # This should always match the max value for the data_explorer_query_result_limit
+  # site setting.
   QUERY_RESULT_MAX_LIMIT = 10000
 
   def self.plugin_name
@@ -117,7 +118,7 @@ after_initialize do
 WITH query AS (
 #{query.sql}
 ) SELECT * FROM query
-LIMIT #{opts[:limit] || DataExplorer::QUERY_RESULT_DEFAULT_LIMIT}
+LIMIT #{opts[:limit] || SiteSetting.data_explorer_query_result_limit}
 SQL
 
           time_start = Time.now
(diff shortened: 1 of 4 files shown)


spec/controllers/queries_controller_spec.rb
```diff
@@ -264,27 +264,19 @@ def run_query(id, params = {})
...
 
         it "should limit the results in JSON response" do
-          begin
-            original_const = DataExplorer::QUERY_RESULT_DEFAULT_LIMIT
-            DataExplorer.send(:remove_const, "QUERY_RESULT_DEFAULT_LIMIT")
-            DataExplorer.const_set("QUERY_RESULT_DEFAULT_LIMIT", 2)
-
-            q = make_query <<~SQL
+          SiteSetting.data_explorer_query_result_limit = 2
+          q = make_query <<~SQL
```
tgxworld, line 268: Minor but since we''re touching this line, I think it is a good opportunity to improve the variable name.

plugin.rb
```diff
@@ -25,7 +25,6 @@
 add_admin_route 'explorer.title', 'explorer'
 
 module ::DataExplorer
-  QUERY_RESULT_DEFAULT_LIMIT = 1000
   QUERY_RESULT_MAX_LIMIT = 10000
```
tgxworld, line 28: Do we still need to keep this constant around?

tgxworld: Ahh it looks we do. Perhaps we can add a comment here to make sure this value and the site setting's value should always be the same?

martin-brennan: Yes it's still used for CSV format:

https://github.com/discourse/discourse-data-explorer/blob/015a35350e6f7375f89b6efeae470edc0df13deb/app/controllers/data_explorer/query_controller.rb#L137-L145

spec/controllers/queries_controller_spec.rb
martin-brennan: Good point, will change to `query`. Man I dislike 1 letter variables unless they are for a loop or something!

plugin.rb
martin-brennan: Your second comment hadn't loaded for me before I sent that, sorry. I will add the comment.

tgxworld (approved):
LGTM :+1:

PR #167, opened 2022-04-05
FIX: Use new properties for pick-files-button

martin-brennan:
Companion commit to core https://github.com/discourse/discourse/pull/16375


tgxworld:
Do we need to add a `.discourse-compatibility` file here since core's API has changed?

martin-brennan:
@tgxworld there already was one, I've updated it now. Can you please double check it, I'm still not totally sure what I am doing with .discourse-compatibility and our internal topic is not straightforward.

tgxworld (approved):
Looks OK but I'm wondering if there are existing tests covering this.

PR #177, opened 2022-06-07
Bookmarkable QueryGroups.

frank3manuel:
Added the ability for non-Admin users to bookmark Queries from inside Group > Reports > Query view.

Limitations: 

1. Because there is no way to view Queries from outside Groups, it is actually the QueryGroup that is being bookmarked. Therefor, the same Query can be bookmarked twice, from different Groups.
2. I haven't been able to turn the bookmark icon blue.
3. DataExplorerQueryGroupBookmarkable.can_see? is not calling guardian.user_is_a_member_of_group? when called from inside the spec, so a workaround was made (DataExplorerQueryGroupBookmarkable.guardian_user_is_a_member_of_group).


spec/lib/data_explorer_query_group_bookmarkable_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+                                    sql: "Not really important",
+                                    user: admin_user) }
+  fab!(:query2) { Fabricate(:query, name: "My Second Query", 
+                                    description: "This is my 2nd query's description.",
+                                    sql: "Not really important",
+                                    user: admin_user) }
+  
+  before do
+    Bookmark.register_bookmarkable(DataExplorerQueryGroupBookmarkable)
+  end
```
tgxworld, line 24: Just curious, do we need to register this again since we've already registered it in `plugin.rb`?

spec/lib/data_explorer_query_group_bookmarkable_spec.rb
```diff
@@ -0,0 +1,153 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe DataExplorerQueryGroupBookmarkable do
+  fab!(:admin_user) { Fabricate(:user) }
```
tgxworld, line 6:
```suggestion
  fab!(:admin_user) { Fabricate(:admin) }
```

Should this be `Fabricate(:admin)` instead?

app/serializers/user_data_explorer_query_group_bookmark_serializer.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class UserDataExplorerQueryGroupBookmarkSerializer < UserBookmarkBaseSerializer
+  attr_reader :data_explorer_query_group
+  attr_reader :data_explorer_query
```
tgxworld, line 5: Are these required when there are methods defined with the same name?

spec/lib/data_explorer_query_group_bookmarkable_spec.rb
martin-brennan: Because the registered bookmarkables are reset between specs. I know it's kind of awkward, not sure how to get around it though:

https://github.com/discourse/discourse/blob/3941bad0751fc2e59576d77bef531b0088571104/spec/rails_helper.rb#L143-L145

martin-brennan: @tgxworld

lib/data_explorer_query_group_bookmarkable.rb
```diff
@@ -0,0 +1,67 @@
...
+  def self.preload_associations
+    [:data_explorer_queries, :groups]
+  end
+
+  def self.list_query(user, guardian)
+    group_ids = user.groups.pluck(:id) 
+    return if group_ids.empty?
+    user.bookmarks_of_type("DataExplorer::QueryGroup")
+    .joins("INNER JOIN data_explorer_query_groups ON data_explorer_query_groups.id = bookmarks.bookmarkable_id
+          AND bookmarks.bookmarkable_type = 'DataExplorer::QueryGroup'")
```
tgxworld, line 21: I don't think this line is necessary since `bookmarks_of_type` is already doing the same thing.

app/serializers/user_data_explorer_query_group_bookmark_serializer.rb
martin-brennan: I think maybe Frank has just seen what I have done and copied that, but what I have done is kind of nonsensical :laughing: I don't think these are required, and I think I can probably remove similar from `UserChatMessageBookmarkSerializer` and `UserPostBookmarkSerializer `. Sorry!

lib/data_explorer_query_group_bookmarkable.rb
```diff
@@ -0,0 +1,67 @@
...
+      .joins("LEFT JOIN data_explorer_queries ON data_explorer_queries.id = data_explorer_query_groups.query_id")
+      .joins("LEFT JOIN groups ON groups.id = data_explorer_query_groups.group_id")
```
tgxworld, line 23: I'm wondering if the two `LEFT JOIN` here are necessary since we don't seem to rely on the joins for further filtering.

assets/javascripts/discourse/controllers/group-reports-show.js
```diff
@@ -35,5 +41,55 @@ export default Ember.Controller.extend({
...
+      return WITH_REMINDER_ICON;
+    }
+    return NO_REMINDER_ICON;
+
+  },
+
+  @discourseComputed("query_group.bookmark")
+  bookmarkClassName(bookmark) {
+    return bookmark
+      ? ["bookmark", "bookmarked"]
```
martin-brennan, line 91: You just need to `.join` these classes, maybe also add a distinct one like `query-group-bookmark` always as well so you can scope some CSS styles like so for the blue icon:

```
.query-group-bookmark {
      &.bookmarked .d-icon {
        color: var(--tertiary);
      }
}
```

app/controllers/data_explorer/query_controller.rb
```diff
@@ -65,7 +65,14 @@ def group_reports_show
 
     respond_to do |format|
       format.json do
-        render_serialized @query, DataExplorer::QuerySerializer, root: 'query'
+        query_group = DataExplorer::QueryGroup.find_by(query_id: @query.id, group_id: @group.id)
+
+        # TODO this has a messy return, a query in a query, and a query_group in a query_group. Fix.
+        render json: {
+          query: serialize_data(@query, DataExplorer::QuerySerializer),
```
martin-brennan, line 72: You just need to add `root: nil` to the end of each of these `serialize_data` calls, otherwise it uses the JSON key as the key for the serialized data, which causes the double up

assets/javascripts/discourse/routes/group-reports-show.js
```diff
@@ -9,15 +9,22 @@ export default DiscourseRoute.extend({
     const group = this.modelFor("group");
     return ajax(`/g/${group.name}/reports/${params.query_id}`)
       .then((response) => {
-        const queryParamInfo = response.query.param_info;
+        // TODO - Messy. Check why DataExplorer::QueryController.group_reports_show
+        //        returns the objects this way.
+        let query = response.query.query;
```
martin-brennan, line 14: See comments I made on the controller.

lib/data_explorer_query_group_bookmarkable.rb
```diff
@@ -0,0 +1,67 @@
...
+      .where("data_explorer_query_groups.group_id IN (?)", group_ids)
+  end
+
+  # Searchable only by data_explorer_queries name and description
+  def self.search_query(bookmarks, query, ts_query, &bookmarkable_search)
+    bookmarkable_search.call(bookmarks, 
+      "data_explorer_queries.name ILIKE :q OR data_explorer_queries.description ILIKE :q")
```
martin-brennan, line 30: I don't think we need to search on description too, just do name for now.

lib/data_explorer_query_group_bookmarkable.rb
martin-brennan: Yeah agree with TGX on the join and `bookmarks_of_type` comments -- if we need to do further filtering later on we can always add these back in.

lib/data_explorer_query_group_bookmarkable.rb
```diff
@@ -0,0 +1,67 @@
...
+  def self.serializer
+    UserDataExplorerQueryGroupBookmarkSerializer
+  end
+
+  def self.preload_associations
+    [:data_explorer_queries, :groups]
+  end
+
+  def self.list_query(user, guardian)
+    group_ids = user.groups.pluck(:id)
```
martin-brennan, line 17: I think `user.visible_groups` would be better here, that also covers the case of a user still being in a group but the group's security settings changing later on.

lib/data_explorer_query_group_bookmarkable.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  def self.reminder_conditions(bookmark)
+    bookmark.bookmarkable.present?
+  end
+
+  def self.can_see?(guardian, bookmark)
+    return false if !bookmark.bookmarkable.group
+
+    # guardian.user_is_a_member_of_group?(bookmark.bookmarkable.group)
```
martin-brennan, line 52: This is happening because `add_to_class` (e.g. `add_to_class(:guardian, :user_is_a_member_of_group?)`) only works if the plugin that is using it is enabled. Just add `SiteSetting.data_explorer_enabled = true` to the `before` block on your spec and this clears it up :) I did this by adding `binding.pry` here and doing:

```ruby
guardian.method(:user_is_a_member_of_group?).source
=> "      klass.public_send(:define_method, attr) do |*args|\n        public_send(hidden_method_name, *args) if plugin.enabled?\n      end\n"
```

There is a lot of magic around Guardian and the plugin API so it's sometimes hard to find things like this.

lib/data_explorer_query_group_bookmarkable.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    # TODO this is a workaround.
+    guardian_user_is_a_member_of_group?(guardian, bookmark.bookmarkable.group)
+    
+  end
+
+  # TODO Lifted almost verbatim from plugin.rb add_to_class(:guardian, :user_can_access_query?)
+  def self.guardian_user_is_a_member_of_group?(guardian, group)
```
martin-brennan, line 60: This can be removed as per other comment.

lib/data_explorer_query_group_bookmarkable.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def self.list_query(user, guardian)
+    group_ids = user.groups.pluck(:id) 
+    return if group_ids.empty?
+    user.bookmarks_of_type("DataExplorer::QueryGroup")
+    .joins("INNER JOIN data_explorer_query_groups ON data_explorer_query_groups.id = bookmarks.bookmarkable_id
+          AND bookmarks.bookmarkable_type = 'DataExplorer::QueryGroup'")
+      .joins("LEFT JOIN data_explorer_queries ON data_explorer_queries.id = data_explorer_query_groups.query_id")
+      .joins("LEFT JOIN groups ON groups.id = data_explorer_query_groups.group_id")
+      .where("data_explorer_query_groups.group_id IN (?)", group_ids)
```
martin-brennan, line 24: I think we only need to conditionally add this if user is not staff -- staff can see every group. Alternatively, we can just not allow admins to bookmark these queries within a group if they are not part of that group. Thoughts @tgxworld ?

martin-brennan (changes requested):
Hey Frank, thanks for the PR, you've done well to follow the bookmarkable pattern and also thank you for commenting your PR with issues you've had and adding an in-depth description to help with review. I've added some comments about the problems you've encountered and also just some general review comments too, if you could please fix those up.

You also just have some linter issues to fix. In discourse core we have a pre-commit hook that catches these before pushing, but we unfortunately don't have that on our plugins.

spec/lib/data_explorer_query_group_bookmarkable_spec.rb
tgxworld: I wonder if we will be able to snapshot the bookmarkable registered at the start of running the suite and then roll back each tests to those bookmarkable at the end of each test.

spec/lib/data_explorer_query_group_bookmarkable_spec.rb
frank3manuel: Ah yes, you're right.

app/serializers/user_data_explorer_query_group_bookmark_serializer.rb
frank3manuel: It's exactly as @martin-brennan says. I wasn't sure if they were actually necessary, but didn't check. I'll confirm they're not necessary and remove them. Thanks!

lib/data_explorer_query_group_bookmarkable.rb
frank3manuel: I was curious about this, but forgot to put a TODO for myself to check. I based this off of ChatMessageBookmarkable, which it turns out is similarly redundant. I'll confirm and update.

lib/data_explorer_query_group_bookmarkable.rb
frank3manuel: Staff can see every group, but it looks like unless the Staff is a member of the group, they can't see the Reports page within the Group. So I think it follows that they shouldn't see the bookmarks if they're not a member of the group either.

For Admin, though, they *can* see the Reports page of a Group they're not a member of, so I can add a condition to `list_query` to consider this. Unless we decide admins shouldn't be able to bookmark from these groups? (They're admins, though. Aren't they usually allowed to do everything while ignoring permissions?)

assets/javascripts/discourse/controllers/group-reports-show.js
frank3manuel: Done.

app/controllers/data_explorer/query_controller.rb
frank3manuel: Done.

assets/javascripts/discourse/routes/group-reports-show.js
frank3manuel: Done.

lib/data_explorer_query_group_bookmarkable.rb
frank3manuel: Done.

lib/data_explorer_query_group_bookmarkable.rb
frank3manuel: Done!

lib/data_explorer_query_group_bookmarkable.rb
frank3manuel: Done.

lib/data_explorer_query_group_bookmarkable.rb
martin-brennan: Ah sorry I meant to say `admin` not `staff`, my bad.

martin-brennan: So yes we need the condition I mentioned but for `admin` users only.

lib/data_explorer_query_group_bookmarkable.rb
martin-brennan: Oh right of course, thanks for that.

lib/data_explorer_query_group_bookmarkable.rb
martin-brennan: Yes makes sense, you are right :+1:

tgxworld: > It seems this INNER JOIN is necessary, since bookmarks_of_type only returns the bookmark record, so there's still a need to join the bookmarkable's table.

Sorry what I meant is that this filter `bookmarks.bookmarkable_type = 'DataExplorer::QueryGroup'"`, is not necessary because "user.bookmarks_of_type" adds the same exact filter. I think the PG planner should be smart enough to only join the relevant `bookmarks` rows to the `data_explorer_query_groups`.

frank3manuel: Oh yeah, you're right, I missed that. Thanks!

lib/data_explorer_query_group_bookmarkable.rb
frank3manuel: Alright then, will apply the condition for admins

frank3manuel: Done!

app/controllers/data_explorer/query_controller.rb
```diff
@@ -65,7 +65,14 @@ def group_reports_show
 
     respond_to do |format|
       format.json do
-        render_serialized @query, DataExplorer::QuerySerializer, root: 'query'
+        query_group = DataExplorer::QueryGroup.find_by(query_id: @query.id, group_id: @group.id)
+
+        # TODO this has a messy return, a query in a query, and a query_group in a query_group. Fix.
```
martin-brennan, line 70: This comment can be removed now.

assets/javascripts/discourse/controllers/group-reports-show.js
```diff
@@ -35,5 +41,55 @@ export default Ember.Controller.extend({
...
+    }
+    return NO_REMINDER_ICON;
+
   },
+
+  @discourseComputed("query_group.bookmark")
+  bookmarkClassName(bookmark) {
+    return bookmark
+      ? ["bookmark", "bookmarked", "query-group-bookmark"].join(" ")
+      : ["bookmark", "query-group-bookmark"];
```
martin-brennan, line 92: Missing a `join` here

assets/javascripts/discourse/routes/group-reports-show.js
```diff
@@ -9,15 +9,20 @@ export default DiscourseRoute.extend({
     const group = this.modelFor("group");
     return ajax(`/g/${group.name}/reports/${params.query_id}`)
       .then((response) => {
-        const queryParamInfo = response.query.param_info;
+        let query = response.query;
+        let query_group = response.query_group;
```
martin-brennan, line 13: nit: We use calemCase for variables in JS

lib/data_explorer_query_group_bookmarkable.rb
```diff
@@ -0,0 +1,53 @@
...
+  def self.serializer
+    UserDataExplorerQueryGroupBookmarkSerializer
+  end
+
+  def self.preload_associations
+    [:data_explorer_queries, :groups]
+  end
+
+  def self.list_query(user, guardian)
+    group_ids = user.visible_groups.pluck(:id)
```
martin-brennan, line 17: An optimization you can do here is to not even fetch the visible_groups if the user is admin:

```
group_ids = []
if !user.admin?
  group_ids = user.visible_groups.pluck(:id)
  return if group_ids.empty?
end
```

martin-brennan (approved):
A few more minor changes but overall this is looking good now! Thanks for the fixes Frank

assets/javascripts/discourse/routes/group-reports-show.js
frank3manuel: Noted. Done.

lib/data_explorer_query_group_bookmarkable.rb
frank3manuel: You're right. Done.

assets/javascripts/discourse/controllers/group-reports-show.js
frank3manuel: You're right. Done.

app/controllers/data_explorer/query_controller.rb
frank3manuel: Done.

PR #185, opened 2022-08-09
FIX: Allow groups to access system queries

frank3manuel:
FIX - allow groups to access system queries (without having to run the query once first)

### DIFF
--- a/app/controllers/data_explorer/query_controller.rb
+++ b/app/controllers/data_explorer/query_controller.rb
@@ -232,7 +232,7 @@ class DataExplorer::QueryController < ::ApplicationController
   end
 
   def set_query
-    @query = DataExplorer::Query.find_by(id: params[:id])
+    @query = DataExplorer::Query.find(params[:id])
     raise Discourse::NotFound unless @query
   end
 end
--- a/spec/requests/query_controller_spec.rb
+++ b/spec/requests/query_controller_spec.rb
@@ -87,6 +87,29 @@ describe DataExplorer::QueryController do
       end
     end
 
+    describe "#update" do
+      fab!(:user2) { Fabricate(:user) }
+      fab!(:group2) { Fabricate(:group, users: [user2]) }
+
+      it "allows group to access system query" do
+        query = DataExplorer::Query.find(-4)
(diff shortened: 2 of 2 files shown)


app/controllers/data_explorer/query_controller.rb
```diff
@@ -232,7 +232,7 @@ def set_group
   end
 
   def set_query
-    @query = DataExplorer::Query.find_by(id: params[:id])
+    @query = DataExplorer::Query.find(params[:id])
```
ZogStriP, line 235: What bug is this fixing? How is it fixing the bug? 🤔

ZogStriP: You should also add a test to ensure we're fixing the bug we say we are fixing and ensure we don't regress at the same time 👍

frank3manuel: > What bug is this fixing? How is it fixing the bug? 🤔

Bug is: Trying to allow a group to access a system query results in a `Discourse::NotFound` unless the query is run first.

Bug report: https://meta.discourse.org/t/open-database-access-katalon-community/231954/6

Cause: 
- System queries don't exist in the database by default
- `update` calls `set_query` before action
- `set_query` searches the database for the system query with `Query.find_by(:id)`, which will not exist by default.
- running system queries first fixes this because `Query.find` is overridden to include system queries (`Queries.default`) in its results, avoiding the `Discourse::NotFound`.

Solution: use the overridden `Query.find` in `set_query` to include system queries in the search, instead of `Query.find_by(:id)`

Test to follow.

ZogStriP: Excellent, that should be in the commit message 👍

frank3manuel: Figuring out how to edit the commit message, as I'm not so well-versed with git and vim. (notice the weird empty force-push below.)

frank3manuel: Commit message updated. Do let me know if I did it wrong.

ZogStriP (approved):

PR #187, opened 2022-09-19
FIX: better handling of edge cases

SamSaffron:
- Require query name is present
- Ensure all routes are treated by default as .json, so errors flow correctly
- Remove superflous save/cancel controls from group settings
- Remove group control when item is destroyed
- Disable editing of query when it is deleted

### DIFF
--- a/assets/javascripts/discourse/templates/admin/plugins-explorer.hbs
+++ b/assets/javascripts/discourse/templates/admin/plugins-explorer.hbs
@@ -88,36 +88,21 @@
               </div>
             {{/if}}
 
-            <div class="pull-left">
-              <div class="groups">
-                <span class="label">{{i18n "explorer.allow_groups"}}</span>
-                <span>
-                  {{multi-select
-                    value=selectedItem.group_ids
-                    content=groupOptions
-                    allowAny=false
-                    onSelect=(action (mut selectedItem.group_ids))
-                  }}
-                </span>
-
-                {{#if runDisabled}}
-                  {{#unless editing}}
-                    <span class="setting-controls">
-                      {{d-button
-                        class="ok"
-                        action=(action "save")
-                        icon="check"
-                      }}
-                      {{d-button
-                        class="cancel"
-                        action=(action "discard")
-                        icon="times"
-                      }}
-                    </span>
-                  {{/unless}}
-                {{/if}}
+            {{#unless selectedItem.destroyed}}
+              <div class="pull-left">
+                <div class="groups">
+                  <span class="label">{{i18n "explorer.allow_groups"}}</span>
+                  <span>
+                    {{multi-select
+                      value=selectedItem.group_ids
+                      content=groupOptions
+                      allowAny=false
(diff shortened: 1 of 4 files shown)


assets/javascripts/discourse/templates/admin/plugins-explorer.hbs
```diff
@@ -88,36 +88,21 @@
...
-                      }}
-                      {{d-button
-                        class="cancel"
-                        action=(action "discard")
-                        icon="times"
-                      }}
-                    </span>
-                  {{/unless}}
-                {{/if}}
+            {{#unless selectedItem.destroyed}}
```
OsamaSayegh, line 91:
```suggestion
            {{#if (not selectedItem.destroyed)}}
```

had to do a double take, is this really you Sam? 😛

assets/javascripts/discourse/templates/admin/plugins-explorer.hbs
```diff
@@ -88,36 +88,21 @@
...
+                      value=selectedItem.group_ids
+                      content=groupOptions
+                      allowAny=false
+                      onSelect=(action (mut selectedItem.group_ids))
+                    }}
+                  </span>
+                </div>
               </div>
-            </div>
+            {{/unless}}
```
OsamaSayegh, line 105:
```suggestion
            {{/if}}
```

spec/requests/query_controller_spec.rb
```diff
@@ -107,6 +107,20 @@ def make_query(sql, opts = {}, group_ids = [])
...
+        query = DataExplorer::Query.find(-4)
+        put "/admin/plugins/explorer/queries/#{query.id}", params: {
+          "query" => {
+            "name" => "",
+          },
+          "id" => query.id }
+
+        expect(response.status).to eq(422)
+        body = JSON.parse(response.body)
+        expect(body["errors"]).to eq(["Name can't be blank"])
```
OsamaSayegh, line 122:
```suggestion
        expect(response.parsed_body["errors"]).to eq(["Name can't be blank"])
```

spec/requests/query_controller_spec.rb
```diff
@@ -304,7 +318,7 @@ def run_query(id, params = {})
             DataExplorer.send(:remove_const, "QUERY_RESULT_MAX_LIMIT")
             DataExplorer.const_set("QUERY_RESULT_MAX_LIMIT", 2)
 
-            ids = Post.order(:id).pluck(:id)
+            _ids = Post.order(:id).pluck(:id)
 
             query = make_query <<~SQL
```
OsamaSayegh, line 323:
```suggestion
            query = make_query <<~SQL
```

seems like this line isn't needed? or at least we don't need to assign it to a variable?

OsamaSayegh (approved):

SamSaffron:
Thanks so much Osama for the review, sorry it has taken so long. not sure how this fell off my radar.

PR #189, opened 2022-10-07
FIX: Errors when running query due to PG template patterns or comments

tgxworld:
Before this fix, the use of PG template patterns containing ":" or the
use of "?" in comments in the SQL will result in an error being raised
because `DB.param_encoder.encode` calls ActiveRecord's `sanitize_sql_array` which is
meant for SQL fragments and not an entire SQL string.

Instead we change data-explorer to use `MiniSql::InlineParamEncoder`
instead which takes into account of template patterns and does not trip
on `?` which is a special param encoding character used by ActiveRecord.

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -76,7 +76,7 @@ after_initialize do
     # Run a data explorer query on the currently connected database.
     #
     # @param [DataExplorer::Query] query the Query object to run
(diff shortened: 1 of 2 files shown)


CvX (approved):

ZogStriP (approved):

tgxworld:
Thank you @CvX and @ZogStriP

PR #191, opened 2022-10-28
FIX: Click not opening query

martin-brennan:
Due to recent core changes for Ember in
https://github.com/discourse/discourse/commit/0221855ba783c6d96d2dbf2ec9f7e92b5297499b and https://github.com/discourse/discourse/commit/952b033165c2b77872d8cc79b0678dc4293c5ee1 the correct way of calling these actions must be observed, otherwise clicking on a query did nothing.

This commit fixes the click and also makes scrollTop work with the new {{on X}} Ember syntax.


nattsw (approved):
Cool, tried it locally and all good.

PR #194, opened 2022-12-12
Upgrade query-row-content to Octane

janzenisaac:


assets/javascripts/discourse/components/query-row-content.js
```diff
@@ -9,62 +9,18 @@ import { get } from "@ember/object";
...
-    const fallback = this.fallbackTemplate;
     const helpers = {
       "icon-or-image": icon_or_image_replacement,
       "category-link": category_badge_replacement,
       reltime: bound_date_replacement,
     };
 
-    const parts = this.columnTemplates.map((t, idx) => {
-      const value = row[idx],
+    this.results = this.args.columnTemplates.map((t, idx) => {
```
davidtaylorhq, line 22: If any arguments change after the initial render, `results` is not going to be recomputed. I guess a general rule-of-thumb to avoid this is "don't use `this.args` inside the constructor".

Previously, `didReceiveAttrs` was re-run on every argument change, so this *might* be a regression (although it obviously depends on precisely how the component is used)

Could we make this a cached getter instead?

```javascript
@cached
get results(){
  return this.args.columnTemplates.map(...
```

janzenisaac: Great tip, thank you. Change added.

davidtaylorhq (approved):

PR #195, opened 2022-12-13
Upgrade share-report to Octane

janzenisaac:


assets/javascripts/discourse/components/share-report.hbs
```diff
@@ -0,0 +1,19 @@
+<div role="button" onclick={{action "open"}}>
```
davidtaylorhq, line 1: Could we make this a real `<button>` or `<DButton>`? `div role=button feels like a bit of a hack.

Either way, let's switch to the `{{on}}` modifier for the click event

```suggestion
<div role="button" {{on "click" this.open}}>
```

assets/javascripts/discourse/components/share-report.hbs
```diff
@@ -0,0 +1,19 @@
...
+  {{group}}
+</div>
+
+{{#if visible}}
+  <div class="popup">
+    <label>{{i18n "explorer.link"}} {{group}}</label>
+    <input type="text" value={{link}}>
+
+    {{d-button
+      action="close"
```
davidtaylorhq, line 12: Pretty sure this'll print a deprecation notice? (if not let me know - we need to fix core)

We should it as a closure action

```suggestion
      action=this.close
```

davidtaylorhq: Oooh I'm confused because there are two `share-report.hbs` templates. The other one looks correct (but should be moved to make it colocated)

assets/javascripts/discourse/components/share-report.hbs
janzenisaac: Sorry for the confusion. As you pointed out there was the old hbs file left in place that I will remove. We need to keep this as is for styling purposes as it is expected to be only displayed as text. I could do something like 

```
  <a href="#" onClick={{this.open}} class="share-report-button">
```

that feels a little more clean.

davidtaylorhq: I see, ok. Happy with whichever you prefer. If we go with the `<a` approach, then we'll have to add `event.preventDefault()` to the handler

assets/javascripts/discourse/templates/components/share-report.hbs
```diff
@@ -1,19 +1,25 @@
...
 
-{{#if visible}}
-  <div class="popup">
-    <label>{{i18n "explorer.link"}} {{group}}</label>
-    <input type="text" value={{link}}>
+  {{#if this.visible}}
+    <div
+      class="popup"
+      {{did-insert this.registerListeners}}
+      {{will-destroy this.unregisterListeners}}
```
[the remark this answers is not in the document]
davidtaylorhq: I see - makes sense :+1:

davidtaylorhq (approved):

PR #201, opened 2022-12-15
Upgrade `data-explorer-bar-chart` to Octane

janzenisaac:


assets/javascripts/discourse/components/data-explorer-bar-chart.js
```diff
@@ -68,31 +69,24 @@ export default Component.extend({
...
-    this._initChart();
-  },
+  @bind
+  initChart(canvas) {
+    loadScript("/javascripts/Chart.min.js").then(() => {
+      const context = canvas.getContext("2d");
+      // eslint-disable-next-line
+      this.chart = new Chart(context, this.config);
+    });
+  }
```
davidtaylorhq, line 81: We could async-ify this:
```suggestion
  @bind
  async initChart(canvas) {
    await loadScript("/javascripts/Chart.min.js");
    const context = canvas.getContext("2d");
    // eslint-disable-next-line
    this.chart = new Chart(context, this.config);
  }
```

davidtaylorhq (approved):

PR #202, opened 2022-12-16
Upgrade `explorer-schema` to Octane

janzenisaac:


assets/javascripts/discourse/controllers/admin-plugins-explorer.js
```diff
@@ -175,10 +175,8 @@ export default Controller.extend({
   },
 
   actions: {
-    dummy() {},
-
-    expandSchema() {
-      this.set("hideSchema", false);
+    updateHideSchema(value) {
+      this.set("hideSchema", value);
```
CvX, line 179: Is `hideSchema` used anywhere still in the controller/its template?

assets/javascripts/discourse/components/explorer-schema.js
```diff
@@ -48,21 +47,24 @@ export default Component.extend({
...
 
-  @discourseComputed("filter")
-  rfilter(filter) {
-    if (!isBlank(filter)) {
-      return new RegExp(filter);
+  get filteredTables() {
+    let tables = [];
+    let filter = this.filter;
+
+    const schema = this.transformedSchema;
```
CvX, line 56: It comes down to personal preference but I usually don't create local vars from `this` props. Makes the code easier to understand w/o context. E.g. seeing `schema` on line 67/68 I'm thinking "what's that?" I scroll up to see it's just `this.transformedSchema`

assets/javascripts/discourse/controllers/admin-plugins-explorer.js
janzenisaac: It is used the in `admin-plugins-explorer` template!

assets/javascripts/discourse/components/explorer-schema.js
janzenisaac: In this case we have to have a variable assignment for `this.filter` as we are manipulating the value without wanting to update the tracked value, but I agree and will update the `schema` const :)

CvX (approved):

PR #203, opened 2022-12-16
Upgrade explorer-schema onetable/enum to Octane

janzenisaac:


assets/javascripts/discourse/components/explorer-schema/one-table.js
```diff
@@ -0,0 +1,17 @@
...
+
+  get styles() {
+    return this.open ? "open" : "";
+  }
+
+  @bind
+  setTableOpen(value) {
+    event.preventDefault();
+    this.open = !value;
+  }
```
CvX, line 16: If the only thing passed to this action is `this.open` I'd rename it to e.g. `toggleOpen` and remove the argument.

assets/javascripts/discourse/components/explorer-schema/one-table.js
```diff
@@ -0,0 +1,17 @@
+import Component from "@glimmer/component";
+import { bind } from "discourse-common/utils/decorators";
+import { tracked } from "@glimmer/tracking";
+
+export default class ExplorerSchemaOneTable extends Component {
```
CvX, line 5:
```suggestion
export default class OneTable extends Component {
```

assets/javascripts/discourse/components/explorer-schema/enum-info.js
```diff
@@ -0,0 +1,15 @@
+import Component from "@glimmer/component";
+
+export default class ParamInput extends Component {
```
CvX, line 3:
```suggestion
export default class EnumInfo extends Component {
```

(or change the file name 😛)

assets/javascripts/discourse/components/explorer-schema/enum-info.js
```diff
@@ -0,0 +1,15 @@
...
+  get enuminfo() {
+    const hash = this.args.col.enum;
+    let result = [];
+    for (let key in hash) {
+      if (!hash.hasOwnProperty(key)) {
+        continue;
+      }
+      result.push({ value: key, name: hash[key] });
+    }
+    return result;
```
CvX, line 13:
```suggestion
    return Object
      .entries(this.args.col.enum)
      .map(([value, name]) => { value, name });
```

janzenisaac: This is causing a https://eslint.org/docs/latest/rules/no-sequences linting error. If I understand this correctly, I don't think we will be able to do the computation of `name` within the comma operator here.

assets/javascripts/discourse/components/explorer-schema/enum-info.js
```diff
@@ -0,0 +1,9 @@
+import Component from "@glimmer/component";
+
+export default class EnumInfo extends Component {
+  get enuminfo() {
+    return Object.entries(this.args.col.enum).map(([value, name]) => {
+      value, name;
+    });
```
CvX, line 7: Whoops, I was missing parens:

```suggestion
    return Object.entries(this.args.col.enum).map(([value, name]) => ({
      value, name;
    }));
```

CvX (approved):

PR #204, opened 2022-12-19
Upgrade `query-result` to Octane

janzenisaac:
I have dropped the unit tests here due to https://github.com/emberjs/ember.js/issues/18877. The tested logic was specifically built with unit tests because we are testing against a `canvas` which gives us no ability to query for dom elements. 

With this in mind what we _could_ do is move (what we can) of the tested logic to exported functions that could then be tested as utility functions. That way they are not component specific... though I don't think that would achieve much as the utility functions it would be testing are very straightforward. So I opted to drop the tests instead.


assets/javascripts/discourse/components/query-result.hbs
```diff
@@ -1,76 +1,77 @@
...
-        action=(action "downloadResultJson")
-        icon="download"
-        label="explorer.download_json"
-        group=group
-      }}
+      <DButton
+        @action={{this.downloadResultJson}}
+        @icon="download"
+        @label="explorer.download_json"
+        @group={{this.group}}
```
davidtaylorhq, line 8: (I realise you didn't introduce it... but...)

I don't recognize the `@group` argument to `DButton`, and I don't see any reference to it in core. Any idea what it does? If not, maybe it should be removed? 🤔

assets/javascripts/discourse/components/query-row-content.js
```diff
@@ -46,23 +57,15 @@ export default class QueryRowContent extends Component {
       }
 
       try {
-        return htmlSafe(
-          (t.template || this.args.fallbackTemplate)(ctx, params)
-        );
+        return htmlSafe((t.template || this.fallbackTemplate)(ctx, params));
```
davidtaylorhq, line 60: Probably something to do in a separate PR... but it would be great if we can get rid of the raw-hbs-based rendering here 🤔

davidtaylorhq (approved):

assets/javascripts/discourse/components/query-result.hbs
janzenisaac: It is not being used, I will drop it.

PR #205, opened 2022-12-20
DEV: Better handling of no results

janzenisaac:
When there were no query results it would throw an error due to `this.resultCount` always passing as it is in the format of

```
"INTEGER - results returned"
```

so we need to grab the first index of the string and check if the integer is great than 0


davidtaylorhq (approved):

tgxworld:
Sorry a couple of things before merging this. 

1.Is there a test we can add for this? 
1. The commit title should follow our commit prefix convention.

PR #206, opened 2022-12-27
DEV: Update params to use previously ran values on reload

janzenisaac:
After running a query with a non-default query param (inserting the new param into the url) we want to have the same params available after reloading the page. To do this we need to pass the updated params back up to the parent due to Octane's one direction data stream. I went over this with @pmusaraj and we both agreed this was extremely difficult to test due to needing to reload the page in a test, so we opted to move forward without one.

- Move param-input tests to a dedicated file


pmusaraj (approved):
Yeah, I don't think it is easy to test this in JS tests. Maybe system tests can be helpful? Worth looking into it in a followup PR.

PR #211, opened 2022-12-29
FIX: Downloading query results

janzenisaac:
We were not passing the correct value from the `query-results-wrapper` to the `query-result` component causing downloads to fail.


janzenisaac:
I will follow up with a test

pmusaraj (approved):

PR #213, opened 2022-12-30
FIX: Group reports downloads

janzenisaac:
We were not passing the correct arguments through in the group reports template causing the download url to default to an admin route. This made downloads unavailable to non-admin users. Fixed by passing the correct values through.


janzenisaac:
I am writing a test for downloads in a separate PR, and will add this case then as well.

gschlager (approved):

PR #218, opened 2023-02-03
DEV: Upgrade `group-reports` to Octane

janzenisaac:


assets/javascripts/discourse/controllers/group-reports-show.js
```diff
@@ -6,93 +6,103 @@ import Bookmark, {
...
+    this.showResults = false;
+
+    ajax(`/g/${this.get("group.name")}/reports/${this.model.id}/run`, {
+      type: "POST",
+      data: {
+        params: JSON.stringify(this.model.params),
+        explain: this.explain,
+      },
+    })
+      .then((result) => {
```
CvX, line 58: You could (if you want 😄) also convert good old promise-based code to async/await

assets/javascripts/discourse/controllers/group-reports-show.js
```diff
@@ -6,93 +6,103 @@ import Bookmark, {
...
+        },
+      }
+    );
+  }
+
+  // This is necessary with glimmer's one way data stream to get the child's
+  // changes of 'params' to bubble up.
+  @action
+  updateParams(identifier, value) {
+    this.set(`model.params.${identifier}`, value);
```
CvX, line 106: Small difference, but let's try not to rely on `.set()` in classes that use `@tracked`:

```suggestion
    this.model.set(`params.${identifier}`, value);
```

(this way the underlying `Controller` class could drop EmberObject dependency and this will still work)

assets/javascripts/discourse/controllers/group-reports-show.js
```diff
@@ -6,93 +6,103 @@ import Bookmark, {
...
+    return openBookmarkModal(
+      this.queryGroupBookmark ||
+        Bookmark.create({
+          bookmarkable_type: "DataExplorer::QueryGroup",
+          bookmarkable_id: this.queryGroup.id,
+          user_id: this.currentUser.id,
+        }),
+      {
+        onAfterSave: (savedData) => {
+          const bookmark = Bookmark.create(savedData);
```
CvX, line 87: Whenever possible it's better to create new records using the data store:

```js
this.store.createRecord("bookmark", savedData);
```

Creating them through the model class/import relies on Discourse-specific hacks to object ownership.

assets/javascripts/discourse/controllers/group-reports-show.js
janzenisaac: added in https://github.com/discourse/discourse-data-explorer/pull/218/commits/69690b905c345a4b6105ea9766392a3239ecbad8

assets/javascripts/discourse/controllers/group-reports-show.js
```diff
@@ -43,34 +44,41 @@ export default class GroupReportsShowController extends Controller {
...
-      .catch((err) => {
-        if (err.jqXHR && err.jqXHR.status === 422 && err.jqXHR.responseJSON) {
-          this.results = err.jqXHR.responseJSON;
-        } else {
-          popupAjaxError(err);
-        }
-      })
-      .finally(() => (this.loading = false));
+      this.results = response;
+      if (!response.success) {
```
janzenisaac, line 65: This feels like a code smell (was existing logic).. if the response is not successful wouldn't we leave the user in a permanent `loading` state since we never make it to setting the loading to false?

assets/javascripts/discourse/controllers/group-reports-show.js
```diff
@@ -84,7 +92,7 @@ export default class GroupReportsShowController extends Controller {
         }),
       {
         onAfterSave: (savedData) => {
-          const bookmark = Bookmark.create(savedData);
+          const bookmark = this.store.createRecord("bookmark", savedData);
```
janzenisaac, line 95: Do I need to save the record?

CvX: Those two forms should be pretty much equivalent. So if no extra save was required before, it shouldn't be required now.

assets/javascripts/discourse/controllers/group-reports-show.js
CvX: Yeah, `this.loading = false` should probably be moved to a `finally` block

assets/javascripts/discourse/controllers/group-reports-show.js
```diff
@@ -6,93 +6,111 @@ import Bookmark, {
...
+
+      this.results = response;
+      if (!response.success) {
+        return;
+      }
+      this.showResults = true;
+    } catch (error) {
+      if (
+        error.jqXHR &&
+        error.jqXHR.status === 422 &&
```
CvX, line 72: 🏌️ ⛳ 

```suggestion
        error.jqXHR?.status === 422 &&
```

assets/javascripts/discourse/controllers/group-reports-show.js
```diff
@@ -6,93 +6,111 @@ import Bookmark, {
...
+    } finally {
+      this.loading = false;
+    }
+  }
+
+  @action
+  toggleBookmark() {
+    return openBookmarkModal(
+      this.queryGroupBookmark ||
+        Bookmark.create({
```
CvX, line 88: the second target for `store.createRecord` 😉

janzenisaac: https://github.com/discourse/discourse-data-explorer/pull/218/commits/ea939f11627ceb33c221874a98140fde6d8f0e01 :)

CvX (approved):

PR #226, opened 2023-02-22
FIX: clear state when reports are changed

lis2:
When show report page is left, results should be cleared to not interfere with next report.

/t/93287


martin-brennan (approved):
Approving this, but this is a perfect chance to introduce system specs into this plugin, since it gives you a very small case that can be easily verified. Every time I fix a plugin now I try add at least one system spec so we can build on this in future.

PR #229, opened 2023-02-27
DEV: Update plugin to match latest guidelines

nbianca:
This commit tries to update the plugin to the latest guidelines, as shown in discourse-plugin-skeleton, which involves moving a lot of the code to dedicated files and use of better namespacing.


markvanlan (approved):
Looks like a bunch of zeitwerk changes, nothing sticks out 👍

PR #232, opened 2023-03-10
FIX: Special attributes not working

janzenisaac:
# Context
Data explorer gives you the ability to use special attributes such as 👇 

> SELECT TIMESTAMP 'yesterday' as reltime$time

# Problem
During the upgrade to ember octane these were neglected and did not work. This PR updates them to working condition.

# Additional
- Drop unused action of `saveDefaults`

### DIFF
--- a/assets/javascripts/discourse/components/query-row-content.js
+++ b/assets/javascripts/discourse/components/query-row-content.js
@@ -1,31 +1,14 @@
 import Component from "@glimmer/component";
-import { categoryLinkHTML } from "discourse/helpers/category-link";
-import { autoUpdatingRelativeAge } from "discourse/lib/formatter";
-import { convertIconClass, iconHTML } from "discourse-common/lib/icon-library";
 import getURL from "discourse-common/lib/get-url";
 import { capitalize } from "@ember/string";
-import { htmlSafe } from "@ember/template";
-import { get } from "@ember/object";
-import { isEmpty } from "@ember/utils";
 import { escapeExpression } from "discourse/lib/utilities";
 import { cached } from "@glimmer/tracking";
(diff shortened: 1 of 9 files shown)


assets/javascripts/discourse/components/query-row-content.js
```diff
@@ -76,32 +53,6 @@ export default class QueryRowContent extends Component {
   }
 }
 
-function icon_or_image_replacement(str, ctx) {
```
janzenisaac, line 79: I moved the helper functions to their dedicated component.

assets/javascripts/discourse/components/result-types/reltime.js
```diff
@@ -0,0 +1,11 @@
+import Component from "@glimmer/component";
+import { autoUpdatingRelativeAge } from "discourse/lib/formatter";
+import { htmlSafe } from "@ember/template";
+
+export default class Reltime extends Component {
+  get boundDateReplacement() {
+    return htmlSafe(
+      autoUpdatingRelativeAge(new Date(this.args.ctx.value), { title: true })
```
davidtaylorhq, line 8: I wonder if we should catch 'invalid date' errors here and return "invalid date" or something? Otherwise a bad query could blow up rendering of the component

davidtaylorhq (approved):

assets/javascripts/discourse/components/result-types/reltime.js
janzenisaac: This is already baked in. I think its a moment.js feature :)

```
-- [params]
-- string :string (set it to a bad value)

SELECT
:string as reltime$updated_at,
```
Result: 
Screenshot 2023-03-10 at 11 41 07 AM

PR #233, opened 2023-03-24
FEATURE: Allow data explorer query result to be sent as recurring PM

dbattersby:
This feature enables admins to create reports automatically based on a recurring schedule.

It introduces a new automation script that includes the new **email_group_user** field added to discourse-automation, along with a **query_id** and **query_params** to pass in parameters to the existing data explorer query.

The output of the report will be sent via pm (as a markdown table) to the recipients entered within the automation script.

The automation (supports individual users, email addresses and groups).

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -81,7 +81,9 @@ after_initialize do
         DiscourseAutomation::Scriptable::RECURRING_DATA_EXPLORER_RESULT_PM,
       ) do
         queries =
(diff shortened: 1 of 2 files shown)


nattsw (approved):

lib/result_to_markdown.rb
nattsw, line 1: Would be nice to have a follow up spec for this file.

jjaffeux:
@dbattersby great job on this 🚀

PR #238, opened 2023-03-29
DEV: Add configurable rate limit for Data Explorer API query runs

Drenmi:
### Background

Data Explorer can run arbitrary SQL queries which can be costly for us if over-used. Because of that we want to add the ability to rate limit the query run endpoint, in particular when requested programmatically using API.

### What is this change?

This introduces a rate limit to the `QueryController#run` endpoint. It heavily leans on the existing `RateLimiter` implementation, and the ability of `ApplicationController` to [turn rate limit exceptions into nicely formatted JSON responses](https://github.com/discourse/discourse/blob/main/app/controllers/application_controller.rb#L180-L199).

The rate limit (per 10 seconds) can be configured through the global setting `max_data_explorer_api_reqs_per_10_seconds`, and defaults to 2.

Handling can be configured through `max_data_explorer_api_req_mode`, and can be set to warn, block, or both warn and block. We will default to `warn` for now and monitor the logs for a while.

### What does a rate limit response look like to end users?

Here's the response body generated from Postman locally:

```json
{
    "errors": [
        "You’ve performed this action too many times. Please wait 5 seconds before trying again."
    ],
    "error_type": "rate_limit",
    "extras": {
        "wait_seconds": 5,
        "time_left": "5 seconds"
    }
}
```

There's also an extra header in the response:

```
Retry-After: 5
```

### Manual verification

- [x] You can run queries in the UI to your heart's content.
- [x] API requests to the run endpoint are rate limited.

### DIFF
--- a/app/controllers/discourse_data_explorer/query_controller.rb
+++ b/app/controllers/discourse_data_explorer/query_controller.rb
@@ -135,6 +135,8 @@ module ::DiscourseDataExplorer
     # explain - string. (Optional - pass explain=true in the request) Postgres query plan, UNIX newlines.
     # rows - array of array of strings. Results of the query. In the same order as 'columns'.
     def run
+      rate_limit_query_runs!
+
       check_xhr unless params[:download]
 
       query = Query.find(params[:id].to_i)
@@ -226,6 +228,22 @@ module ::DiscourseDataExplorer
     end
 
     private
+
+    def rate_limit_query_runs!
+      return if !is_api? && !is_user_api?
+
+      RateLimiter.new(
+        nil,
+        "api-query-run-10-sec",
+        GlobalSetting.max_data_explorer_api_reqs_per_10_seconds,
+        10.seconds,
+      ).performed!
+    rescue RateLimiter::LimitExceeded => e
+      if GlobalSetting.max_data_explorer_api_req_mode.include?("warn")
+        Discourse.warn("Query run 10 second rate limit exceeded", query_id: params[:id])
+      end
+      raise e if GlobalSetting.max_data_explorer_api_req_mode.include?("block")
+    end
 
     def set_group
       @group = Group.find_by(name: params["group_name"])
--- a/plugin.rb
+++ b/plugin.rb
@@ -38,6 +38,14 @@ after_initialize do
   require_relative "lib/discourse_data_explorer/queries"
   require_relative "lib/discourse_data_explorer/query_group_bookmarkable"
 
+  GlobalSetting.add_default(:max_data_explorer_api_reqs_per_10_seconds, 2)
+
+  # Available options:
+  #   - warn
+  #   - warn+block
+  #   - block
+  GlobalSetting.add_default(:max_data_explorer_api_req_mode, "warn")
+
   add_to_class(:guardian, :user_is_a_member_of_group?) do |group|
     return false if !current_user
     return true if current_user.admin?
--- a/spec/requests/query_controller_spec.rb
+++ b/spec/requests/query_controller_spec.rb
@@ -170,6 +170,97 @@ describe DiscourseDataExplorer::QueryController do
         expect(response_json["errors"].first).to match(/ValidationError/)
       end
 
+      context "when rate limited" do
+        def unlimited_request(query_id, headers = {})
+          post "/admin/plugins/explorer/queries/#{query_id}/run.json",
+               params: {
+                 params: {}.to_json,
+               },
+               headers: headers
+
+          expect(response.status).to eq(200)
+        end
+
+        def limited_request(query_id, headers = {})
+          post "/admin/plugins/explorer/queries/#{query_id}/run.json",
+               params: {
+                 params: {}.to_json,
+               },
+               headers: headers
+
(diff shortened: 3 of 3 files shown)


plugin.rb
```diff
@@ -38,6 +38,8 @@ module ::DiscourseDataExplorer
   require_relative "lib/discourse_data_explorer/queries"
   require_relative "lib/discourse_data_explorer/query_group_bookmarkable"
 
+  GlobalSetting.add_default(:max_data_explorer_api_reqs_per_10_seconds, 2)
```
Drenmi, line 41: Need some validation whether this is a good way to add a global setting from a plugin. I verified that this works, and can be overridden e.g. from `discourse_defaults.conf`, so functionally it does the trick.

app/controllers/discourse_data_explorer/query_controller.rb
```diff
@@ -135,6 +135,8 @@ def schema
     # explain - string. (Optional - pass explain=true in the request) Postgres query plan, UNIX newlines.
     # rows - array of array of strings. Results of the query. In the same order as 'columns'.
     def run
+      rate_limit_query_runs!
```
Drenmi, line 138: This can raise a `RateLimiter::LimitExceeded` error which will be caught and turned into a nice JSON payload with extra headers [here](https://github.com/discourse/discourse/blob/main/app/controllers/application_controller.rb#L180-L199).

plugin.rb
martin-brennan: As noted on the call, yes it's all good:

https://github.com/discourse/discourse/blob/3a9a534881135ac8928266b1d02bc679f2b49b31/plugins/chat/plugin.rb#L28-L28

app/controllers/discourse_data_explorer/query_controller.rb
```diff
@@ -227,6 +229,22 @@ def run
 
     private
 
+    def rate_limit_query_runs!
+      return if !is_api? && !is_user_api?
```
martin-brennan, line 233: I am guessing these come from `DefaultCurrentUserProvider`? I wonder if we can do something like add a method called `is_api_call?` that covers both of these? Just a suggestion, it is fine as is as well, just could potentially be confusing for future engineers since you have to know this is in the current user provider

spec/requests/query_controller_spec.rb
```diff
@@ -170,6 +170,94 @@ def run_query(id, params = {})
...
+               headers: headers
+
+          expect(response.status).to eq(429)
+          expect(response.parsed_body["extras"]).to eq(
+            { "wait_seconds" => 9, "time_left" => "9 seconds" },
+          )
+        end
+
+        it "limits query runs from API when using block mode" do
+          GlobalSetting.stubs(:max_data_explorer_api_reqs_per_10_seconds).returns(1)
```
martin-brennan, line 198: We have a helper for this stubbing, see:

https://github.com/discourse/discourse/blob/3a9a534881135ac8928266b1d02bc679f2b49b31/spec/lib/global_path_spec.rb#L30-L44

https://github.com/discourse/discourse/blob/3a9a534881135ac8928266b1d02bc679f2b49b31/spec/rails_helper.rb#L478-L484

app/controllers/discourse_data_explorer/query_controller.rb
```diff
@@ -227,6 +229,22 @@ def run
...
+
+      RateLimiter.new(
+        nil,
+        "api-query-run-10-sec",
+        GlobalSetting.max_data_explorer_api_reqs_per_10_seconds,
+        10.seconds,
+      ).performed!
+    rescue RateLimiter::LimitExceeded => e
+      if GlobalSetting.max_data_explorer_api_req_mode.include?("warn")
+        Discourse.warn("Query run 10 second rate limit exceeded")
```
martin-brennan, line 243: I wonder if we interpolate the query ID into this for extra data? You can pass an additional hash to `Discourse.warn`:

https://github.com/discourse/discourse/blob/3a9a534881135ac8928266b1d02bc679f2b49b31/app/models/optimized_image.rb#L347-L353

martin-brennan (approved):
Looks nice, approving because it works as is but just have a few comments/suggestions :)

app/controllers/discourse_data_explorer/query_controller.rb
Drenmi: I was thinking along the same lines. It'd need to be a change in core, but at the same time I think maybe this should accompany some bigger work around API stuff. I know TGX has been thinking about this a lot. 🙂

app/controllers/discourse_data_explorer/query_controller.rb
Drenmi: Good idea! I saw it accepts a hash, but was on the fence since the query that triggers the rate limit is just "the straw that breaks the camel's back", but can probably still be useful for advocates. 🙂

app/controllers/discourse_data_explorer/query_controller.rb
martin-brennan: Sure no problem this is fine as is :)

PR #239, opened 2023-04-05
[FIX] The user_list parameter returned user objects instead of their IDs

Canapin:
The user_list parameter returned the complete User objects instead of users' IDs, making the query fail.


janzenisaac:
Looks good, I think we need to tweak the pr name / description. The user_list parameter was returning a list of usernames instead of ids, eg.

```
{"user_ids":"foo_username_here"}
```

So maybe,  _FIX: The user_list parameter returned a list of usernames instead of ids_

janzenisaac (approved):

PR #241, opened 2023-04-17
FIX: Reduce issues when debouncing schema search

pmusaraj:
Previously the debounced value was both setting the filter value and updating the input's text value. This causes visible UI issues, because the debounce updating the input's text value would sometimes reset it, especially if/when typing quickly.

This separates the `filter` onChange event from the input value.

This PR also uses sticky positioning for the search form, so that it stays visible even when scrolling the list of schema tables.

screencast 2023-04-17 18-42-18

### DIFF
--- a/assets/javascripts/discourse/components/explorer-schema.hbs
+++ b/assets/javascripts/discourse/components/explorer-schema.hbs
@@ -7,7 +7,10 @@
 {{else}}
   <div class="schema">
     <div class="schema-search inline-form full-width">
-      <Input {{on "input" (action "filterChanged" value="target.value")}} />
+      <input
+        type="text"
+        {{on "input" (action "filterChanged" value="target.value")}}
+      />
       <DButton
         @action={{this.collapseSchema}}
         @icon="chevron-right"
--- a/assets/javascripts/discourse/components/explorer-schema.hbs
+++ b/assets/javascripts/discourse/components/explorer-schema.hbs
@@ -7,11 +7,7 @@
 {{else}}
   <div class="schema">
     <div class="schema-search inline-form full-width">
-      <TextField
-        @value={{this.filterInput}}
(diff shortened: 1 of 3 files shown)


assets/javascripts/discourse/components/explorer-schema.hbs
```diff
@@ -4,13 +4,11 @@
...
-{{/if}}
-
-<div class="schema">
-  <div class={{if this.hideSchema "hidden"}}>
+{{else}}
+  <div class="schema">
     <div class="schema-search inline-form full-width">
       <TextField
-        @value={{this.filter}}
+        @value={{this.filterInput}}
```
CvX, line 11: If `filterInput` isn't read anywhere, and its value is set only by this textfield then I think it's not needed at all?

pmusaraj: Good catch. Our `TextField` wrapper needed it, but `Input` doesn't need it, and using `Input` is cleaner (fewer debounces as well).

janzenisaac (approved):
Awesome!

assets/javascripts/discourse/components/explorer-schema.hbs
```diff
@@ -4,31 +4,25 @@
...
-  <div class={{if this.hideSchema "hidden"}}>
+{{else}}
+  <div class="schema">
     <div class="schema-search inline-form full-width">
-      <TextField
-        @value={{this.filter}}
-        @onChange={{this.filterChanged}}
-        @placeholderKey="explorer.schema.filter"
-      />
+      <Input {{on "input" (action "filterChanged" value="target.value")}} />
```
CvX, line 10: To further simplify things, a standard html element can be used here now, I think:

```suggestion
      <input {{on "input" (action "filterChanged" value="target.value")}} />
```

PR #245, opened 2023-05-23
UX: fix parameters inputs spacing

Canapin:
Fix the spacing between parameters inputs.

Before:
image

After (I forgot to add a select kit on the first screenshot, but that doesn't really matter here):
image

I removed some arbitrary widths that didn't seem necessary and also center-aligned content.
Elements will go on multiple lines if the container is too narrow.
If a select-kit is populated with a long string, it will overflow on mobile, but I don't think it's an issue since data explorer isn't exactly made for mobile (it already has a select-kit overflow I didn't bother fixing).

Besides this, the fields will be properly displayed on mobile in normal use anyway.

### DIFF
--- a/assets/stylesheets/explorer.scss
+++ b/assets/stylesheets/explorer.scss
@@ -457,4 +457,3 @@ table.group-reports {
     color: var(--tertiary);
   }
 }
-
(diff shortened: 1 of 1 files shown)


pmusaraj (approved):
Looks better!

PR #250, opened 2023-07-04
FIX: Scheduled PM with Data Explorer Results not sending PM

dbattersby:
This PR fixes 2 issues that were picked up by users for the Scheduled Data Explorer Report automation script and a couple of small improvements to better match the format of manual data explorer query results.

The first issue is that within `result_to_markdown` the `colrender` contains null values, and we are currently checking length (previously treated as a packed array but it is actually a sparse array). Therefore we can check if the current index of the array is null rather than checking the size of the array.

The second issue addresses the blank `query_params` field. When the data explorer script does not require any params to be passed in via the automation script then it will have a `nil` value, however it should be defaulted to `{}` within the plugin.

To improve formatting the markdown table for PMs is now aligned to left and where values are substituted (for example user_id becomes username) we then include the id within brackets, for example:
`user_id` becomes `username (user_id)`

I also added introduced some tests for the result_to_markdown class to ensure the data is returned in markdown format and uses the enriched data from `DataExplorer.add_extra_data`.

[Meta Topic](https://meta.discourse.org/t/automation-script-schedule-pm-with-data-explorer-results-not-sending-pms/269135)

### DIFF
--- a/spec/result_to_markdown_spec.rb
+++ b/spec/result_to_markdown_spec.rb
@@ -4,7 +4,7 @@ describe DiscourseDataExplorer::ResultToMarkdown do
   fab!(:user) { Fabricate(:user) }
   fab!(:post) { Fabricate(:post) }
   fab!(:query) { DiscourseDataExplorer::Query.find(-1) }
-  
+
   let(:query_params) { [{ from_days_ago: 0 }, { duration_days: 15 }] }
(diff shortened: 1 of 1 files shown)


martin-brennan (approved):
Looks great and really well done PR description too :clap:

PR #260, opened 2023-10-31
FEATURE: Add JSON result type component

martin-brennan:
If a column is `payload` or contains `_payload` it will be assumed
it has JSON data in it, then we will show the truncated JSON in the
result column with a button to show the full-screen formatted
JSON using our full-screen code viewer.

image
image

### DIFF
--- a/assets/javascripts/discourse/components/result-types/json.hbs
+++ b/assets/javascripts/discourse/components/result-types/json.hbs
@@ -1,0 +1,9 @@
+<div class="result-json">
+  <div class="result-json-value">{{@ctx.value}}</div>
+  <DButton
+    @class="result-json-button"
+    @action={{action "viewJson"}}
+    @icon="ellipsis-h"
+    @title="explorer.view_json"
+  />
+</div>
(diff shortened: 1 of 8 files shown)


assets/javascripts/discourse/components/result-types/json.js
```diff
@@ -0,0 +1,35 @@
...
+    }
+  }
+
+  get elidedValue() {
+    return this.args.ctx.value.slice(0, 50) + "...";
+  }
+
+  @action
+  viewJson() {
+    this.modal.show(FullscreenCodeModal, {
```
jjaffeux, line 26: We should avoid doing this, it's considering any modal we write in core as public API.

Not sure what platform would prefer here instead, worth discussing it.

martin-brennan: Fair enough, but AFAIK we don't have any patterns for this yet and it would be nice to not hold this up until TGX is back from holiday. This modal isn't likely to change anytime soon.

martin-brennan: I will make a topic about it too.

assets/javascripts/discourse/components/result-types/json.hbs
```diff
@@ -0,0 +1,9 @@
+<div class="result-json">
+  <div class="result-json-value">{{@ctx.value}}</div>
+  <DButton
+    @class="result-json-button"
```
jjaffeux, line 4:
```suggestion
    class="result-json-button"
```

jjaffeux (approved):

PR #274, opened 2024-01-02
FIX: Continue to send the PM to others when there is a group or user that does not exist

nattsw:
Currently there is an error at `report_generator.rb:14:in `generate'` 

```
Job exception: undefined method `empty?' for nil:NilClass
```

which happens when an automation is created to send to a user/group that does not exist. This lets the function continue and continue sending the automated PM to the rest of the users/groups.

### DIFF
--- a/lib/report_generator.rb
+++ b/lib/report_generator.rb
@@ -10,8 +10,12 @@ module ::DiscourseDataExplorer
       query = DiscourseDataExplorer::Query.find(query_id)
       return [] unless query
 
+      creator = User.find(@creator_user_id)
+      return [] unless Guardian.new(creator).can_send_private_messages?
+
+      return [] if recipients.empty?
+
       usernames = filter_recipients_by_query_access(recipients, query)
-      return [] if usernames.empty?
       params = params_to_hash(query_params)
 
       result = DataExplorer.run_query(query, params)
@@ -23,17 +27,15 @@ module ::DiscourseDataExplorer
     end
 
     def filter_recipients_by_query_access(recipients, query)
-      return [] if recipients.empty?
-      creator = User.find(@creator_user_id)
-      return [] unless Guardian.new(creator).can_send_private_messages?
-
       recipients.reduce([]) do |names, recipient|
         if (group = Group.find_by(name: recipient))
-          next names unless query.query_groups.exists?(group_id: group.id)
-          next names.concat group.users.pluck(:username)
+          names unless query.query_groups.exists?(group_id: group.id)
+          names.concat group.users.pluck(:username)
         elsif (user = User.find_by(username: recipient))
-          next names unless Guardian.new(user).user_can_access_query?(query)
-          next names << recipient
+          names unless Guardian.new(user).user_can_access_query?(query)
+          names << recipient
+        else
+          names
         end
       end
     end
--- a/lib/report_generator.rb
+++ b/lib/report_generator.rb
@@ -9,11 +9,10 @@ module ::DiscourseDataExplorer
     def generate(query_id, query_params, recipients)
       query = DiscourseDataExplorer::Query.find(query_id)
       return [] unless query
-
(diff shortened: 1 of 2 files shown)


nattsw:
Checking the failure

lib/report_generator.rb
```diff
@@ -10,8 +10,12 @@ def generate(query_id, query_params, recipients)
       query = DiscourseDataExplorer::Query.find(query_id)
       return [] unless query
 
+      creator = User.find(@creator_user_id)
```
tgxworld, line 13: Do we want this to raise an error if the user cannot be found? If not, it is better to use `find_by` instead.

```suggestion
      creator = User.find_by(id: @creator_user_id)
```

lib/report_generator.rb
```diff
@@ -10,8 +10,12 @@ def generate(query_id, query_params, recipients)
...
+      return [] unless Guardian.new(creator).can_send_private_messages?
+
+      return [] if recipients.empty?
```
tgxworld, line 16:
```suggestion
      return [] if recipients.empty?

      creator = User.find(@creator_user_id)
      return [] unless Guardian.new(creator).can_send_private_messages?
```

Slightly more efficient since we don't run queries unnecessarily when `recipients` is empty.

lib/report_generator.rb
```diff
@@ -23,18 +27,17 @@ def generate(query_id, query_params, recipients)
     end
 
     def filter_recipients_by_query_access(recipients, query)
```
tgxworld, line 29: Minor but I think this method should probably be private now that it doesn't do the basic guardian checks.

lib/report_generator.rb
```diff
@@ -10,8 +10,12 @@ def generate(query_id, query_params, recipients)
       query = DiscourseDataExplorer::Query.find(query_id)
       return [] unless query
 
+      creator = User.find_by(id: @creator_user_id)
+      return [] unless Guardian.new(creator).can_send_private_messages?
+
+      return [] if recipients.empty?
```
tgxworld, line 16:
```suggestion
      return [] if recipients.empty?

      creator = User.find_by(id: @creator_user_id)
      return [] unless Guardian.new(creator).can_send_private_messages?
```

Minor but this order will avoid running a DB query when `recipients` is empty.

lib/report_generator.rb
```diff
@@ -72,5 +59,22 @@ def build_report_pms(query, table = "", usernames = [])
...
+        if (group = Group.find_by(name: recipient)) &&
+             (
+               group.id == Group::AUTO_GROUPS[:admins] ||
+                 query.query_groups.exists?(group_id: group.id)
+             )
+          next names.concat group.users.pluck(:username)
+        elsif (user = User.find_by(username: recipient))
+          next names << recipient if Guardian.new(user).user_can_access_query?(query)
+        end
+        next names
```
tgxworld, line 76:
```suggestion
      recipients.reduce([]) do |names, recipient|
        if (group = Group.find_by(name: recipient)) &&
             (
               group.id == Group::AUTO_GROUPS[:admins] ||
                 query.query_groups.exists?(group_id: group.id)
             )
          names.concat group.users.pluck(:username)
        elsif (user = User.find_by(username: recipient))
          names << recipient if Guardian.new(user).user_can_access_query?(query)
        else
          names
        end
```

tgxworld (approved):

lib/report_generator.rb
nattsw: Your line 74 will result in a `names` becoming `nil` if the condition does not satisfy.

nattsw: Haha saw the edit. I will leave it as is.

tgxworld: Yea the key with `reduce` is that you have to return the state at the end of every iteration. I think that is what was causing the problem in the first place.

PR #275, opened 2024-01-12
FEATURE: Allow topic URL without post number for post_id param

martin-brennan:
This commit allows base topic URLs (e.g. https://meta.discourse.org/t/hide-text-in-text-select-popup-menu-feedback/287656)
to be used by the post_id parameter type. We just assume in this case
that the post_id for the param is 1.

### DIFF
--- a/lib/discourse_data_explorer/parameter.rb
+++ b/lib/discourse_data_explorer/parameter.rb
@@ -67,6 +67,46 @@ module ::DiscourseDataExplorer
       @type_aliases ||= { integer: :int, text: :string, timestamp: :datetime }
     end
 
+    def self.create_from_sql(sql, opts = {})
+      in_params = false
+      ret_params = []
+      sql.lines.find do |line|
+        line.chomp!
+
+        if in_params
+          # -- (ident) :(ident) (= (ident))?
+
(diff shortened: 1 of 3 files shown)


lib/discourse_data_explorer/parameter.rb
```diff
@@ -67,6 +67,46 @@ def self.type_aliases
       @type_aliases ||= { integer: :int, text: :string, timestamp: :datetime }
     end
 
+    def self.create_from_sql(sql, opts = {})
```
martin-brennan, line 70: This was just moved around so the class + instance methods are grouped properly

lib/discourse_data_explorer/parameter.rb
```diff
@@ -173,9 +205,13 @@ def invalid_format(string, msg = nil)
             invalid_format string, "The user named #{string} was not found"
           end
         elsif type == :post_id
-          if string =~ %r{(\d+)/(\d+)(\?u=.*)?$}
+          if string =~ %r{/t/[^/]+/(\d+)(\?u=.*)?$}
```
martin-brennan, line 208: Main change is here

tgxworld (approved):

martin-brennan:
Thanks @tgxworld !

PR #284, opened 2024-03-22
FIX: send single report PM to groups

dbattersby:
The main change here is that we now send a single PM to a group rather than individual PMs to each group member.

We are also now setting email recipients correctly as `target_emails`, support was added for emails in [this change](https://github.com/discourse/discourse-automation/pull/262).

### DIFF
--- a/lib/report_generator.rb
+++ b/lib/report_generator.rb
@@ -62,13 +62,13 @@ module ::DiscourseDataExplorer
         .uniq
         .reduce([]) do |names, recipient|
           if (group = Group.find_by(name: recipient)) &&
-            (
-              group.id == Group::AUTO_GROUPS[:admins] ||
-              query.query_groups.exists?(group_id: group.id)
-            )
+               (
+                 group.id == Group::AUTO_GROUPS[:admins] ||
+                   query.query_groups.exists?(group_id: group.id)
+               )
             names << [recipient, "group"]
           elsif (user = User.find_by(username: recipient)) &&
-            Guardian.new(user).user_can_access_query?(query)
+                Guardian.new(user).user_can_access_query?(query)
             names << [recipient, "user"]
           elsif Email.is_valid?(recipient)
             names << [recipient, "email"]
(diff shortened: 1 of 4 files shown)


lib/report_generator.rb
```diff
@@ -2,30 +2,24 @@
 
 module ::DiscourseDataExplorer
   class ReportGenerator
-    def initialize(creator_user_id)
```
dbattersby, line 5: No longer needed as the PM is sent by the system user anyway.

lib/report_generator.rb
```diff
@@ -2,30 +2,24 @@
...
-    end
-
-    def generate(query_id, query_params, recipients)
+    def self.generate(query_id, query_params, recipients)
       query = DiscourseDataExplorer::Query.find(query_id)
-      return [] unless query
-      return [] if recipients.empty?
-
-      creator = User.find_by(id: @creator_user_id)
-      return [] unless Guardian.new(creator).can_send_private_messages?
```
dbattersby, line 15: Not needed as the PM is sent by the system user.

lib/report_generator.rb
```diff
@@ -60,23 +57,24 @@ def build_report_pms(query, table = "", usernames = [])
...
-             )
-          names.concat group.users.pluck(:username)
-        elsif (user = User.find_by(username: recipient)) &&
-              Guardian.new(user).user_can_access_query?(query)
-          names << recipient
+    def self.filter_recipients_by_query_access(recipients, query)
+      recipients
+        .uniq
+        .reduce([]) do |names, recipient|
+          if (group = Group.find_by(name: recipient)) &&
```
jjaffeux, line 64: This one and others (like User.find_by) feels like an N+1 no? (I see it was already present before)

dbattersby: My expectation is that there will likely only be 3-4 recipients entered, and the script is only used by site admins. But I'm open to refactoring this if it's a concern that it could be misused.

Multiple recipient types are here so an alternative might be to do `.where` against groups and users, then loop the results against Guardian and remove those that fail.

jjaffeux: Yes if you can try to optimize this a little bit that would be great please

jjaffeux (approved):
Thanks!

PR #291, opened 2024-05-21
FIX: Bookmarking group reports

martin-brennan:
Since https://github.com/discourse/discourse/commit/67a8080e338aa62afbd02efc98178b4a67751775
in core, the functionality to bookmark a report from the group
Reports tab has been broken. This commit fixes the issue and adds
system spec coverage to prevent regression.

### DIFF
--- a/spec/system/bookmark_spec.rb
+++ b/spec/system/bookmark_spec.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+describe "Bookmarking reports attached to a group", type: :system, js: true do
+  fab!(:current_user) { Fabricate(:admin) }
+  fab!(:query_1) do
+    Fabricate(
+      :query,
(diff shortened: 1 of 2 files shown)


ZogStriP (approved):

spec/system/bookmark_spec.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+describe "Bookmarking reports attached to a group", type: :system, js: true do
```
CvX, line 3: 🏌️ all system specs have js! 😃 

```suggestion
describe "Bookmarking reports attached to a group", type: :system do
```

martin-brennan: Lol this is my snippet, I will update it to remove this :)

PR #128, opened 2021-07-28
Add Data Explorer Params to the URL when user press run

rngus2344:

PR #298, opened 2024-07-05
FEATURE: Add Data Explorer Params to the URL for group queries

Lhcfl:
We have introduced a URL param mechanism for data explorer queries for administrators in https://github.com/discourse/discourse-data-explorer/pull/128/ However, for data explorer queries in group page, URL params are not yet introduced.

This PR introduces this missing piece.

after:

image

Related meta topic: https://meta.discourse.org/t/populate-data-explorer-params-with-url-params/169404/8

### DIFF
--- a/test/javascripts/acceptance/param-input-test.js
+++ b/test/javascripts/acceptance/param-input-test.js
@@ -304,7 +304,18 @@ acceptance("Data Explorer Plugin | Param Input", function (needs) {
     await fillIn(".query-params input", monthsAgoValue);
     await click("form.query-run button");
 
-    const searchParams = new URLSearchParams(currentURL());
+    const searchParams = new URLSearchParams(currentURL().split("?")[1]);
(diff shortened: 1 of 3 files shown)


test/javascripts/acceptance/param-input-test.js
```diff
@@ -304,7 +304,18 @@ acceptance("Data Explorer Plugin | Param Input", function (needs) {
     await fillIn(".query-params input", monthsAgoValue);
     await click("form.query-run button");
 
-    const searchParams = new URLSearchParams(currentURL());
```
Lhcfl, line 307: About this: the result of `Array.from(new URLSearchParams('/something?id=1&params={"a":1}').entries())` is actually `[ ['/something?id', '1'], ['params', '{"a":1}']]`

So the original logic is actually misleading. :face_holding_back_tears:

ZogStriP (approved):

PR #303, opened 2024-08-02
FIX: Create parameter input boxes after save

Lhcfl:
What does this fix?
===================

When creating a data explorer query that includes parameters, the parameter input boxes don’t display until the page is refreshed.

After this commit, when the query is saved, the input boxes will appear.

Screen Record
===========

https://github.com/user-attachments/assets/d9c4a412-9150-4c7e-a19f-85d6fd1c313b

ref: t/113297


ZogStriP (approved):
LGTM 👍 

Any chances you could add a spec to ensure this behavior doesn't break in the future?

Lhcfl:
added!

Lhcfl:
@ZogStriP Could you take another look at how this test is written?

PR #306, opened 2024-08-12
UX: Using CategoryChooser for param param_input

Lhcfl:
This change changes the category_id input to use CategoryChooser instead of a bare text input to improve the user experience.

Related meta link: https://meta.discourse.org/t/wishlist-param-dropdown-for-data-explorer-query/253883/28

## Screenshot:

image


ZogStriP (approved):
Nice 😍

PR #310, opened 2024-08-20
DEV: Use Category's async method in param-input

Lhcfl:
This commit refactors the category id normalization of param-input to use `Category.asyncFindBySlugPath`.

Now, the category id input will use an async query when the default/initial value is category slug.

#### When an incorrect default category_id is provided

image

#### A loading category
image


Lhcfl:
Thanks for your help @nat !

PR #315, opened 2024-08-21
UX: Use GroupChooser in `group_id` param input

Lhcfl:
This commit uses GroupChooser as the input for param input of type group_id. Meanwhile, it improves invalid group validation and semantic error prompts.

## Screenshot:

image


assets/javascripts/discourse/components/param-input/group-list-input.gjs
```diff
@@ -2,13 +2,21 @@ import Component from "@glimmer/component";
 import { inject as service } from "@ember/service";
 import GroupChooser from "select-kit/components/group-chooser";
 
-export default class GroupListInput extends Component {
+export default class GroupInput extends Component {
```
jjaffeux, line 5: the filename is still group-list-input

Lhcfl: wow that was my oversight!

assets/javascripts/discourse/components/param-input-form.gjs
```diff
@@ -158,6 +134,9 @@ export default class ParamInputForm extends Component {
   form = null;
 
   promiseNormalizations = [];
+  formLoaded = new Promise((res) => {
+    this.__form_load_callback = res;
+  });
```
jjaffeux, line 139: Is this used to show existing errors of previous params right when loading the form without any user interaction?

Lhcfl: Yes. It may can also be done directly with setTimeout 0, I'm not sure which is better

jjaffeux (approved):
Ok that looks good to me, ultimately I want to support this use case in form kit directly. The `validateOn` is the thing Im not super happy in form-kit atm, it's buggy and doesnt do exactly what it says and should also support some kind of "live" mode, where it does validate the initial state. I will try to carve some time to do this in the future.

PR #318, opened 2024-08-26
FEATURE: Allows CSV file result to be attached in automated PMs

Lhcfl:
This commit adds an optional setting that allows to attach query results in CSV format as a file to PMs sent by Data Explorer's automation scripts.

meta topic: https://meta.discourse.org/t/turn-data-explorer-query-results-into-csv-to-attach-to-discourse-automated-emails/267529

## Screenshots

image
image

### DIFF
--- a/app/controllers/discourse_data_explorer/query_controller.rb
+++ b/app/controllers/discourse_data_explorer/query_controller.rb
@@ -179,49 +179,25 @@ module ::DiscourseDataExplorer
 
         render json: { success: false, errors: [err_msg] }, status: 422
       else
-        pg_result = result[:pg_result]
-        cols = pg_result.fields
+        filename =
+          "attachment; filename=#{query.slug}@#{Slug.for(Discourse.current_hostname, "discourse")}-#{Date.today}.dcqresult"
+
         respond_to do |format|
           format.json do
-            if params[:download]
-              response.headers[
-                "Content-Disposition"
-              ] = "attachment; filename=#{query.slug}@#{Slug.for(Discourse.current_hostname, "discourse")}-#{Date.today}.dcqresult.json"
-            end
-            json = {
-              success: true,
-              errors: [],
-              duration: (result[:duration_secs].to_f * 1000).round(1),
-              result_count: pg_result.values.length || 0,
-              params: query_params,
-              columns: cols,
-              default_limit: SiteSetting.data_explorer_query_result_limit,
-            }
-            json[:explain] = result[:explain] if opts[:explain]
+            response.headers["Content-Disposition"] = "#{filename}.json" if params[:download]
 
-            if !params[:download]
-              relations, colrender = DataExplorer.add_extra_data(pg_result)
-              json[:relations] = relations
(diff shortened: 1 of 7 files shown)


lib/report_generator.rb
```diff
@@ -9,13 +9,13 @@ def self.generate(query_id, query_params, recipients, opts = {})
       recipients = filter_recipients_by_query_access(recipients, query)
       params = params_to_hash(query_params)
 
-      result = DataExplorer.run_query(query, params)[:pg_result]
+      result = DataExplorer.run_query(query, params)
```
Drenmi, line 12: I'm thinking if we can promote the return value from a hash to a class or struct instead. Then we could implement for example `result.empty?` and use that on L:15. It could also implement `result.convert` which could be delegated to the new converter class. WDYT?

lib/report_generator.rb
```diff
@@ -40,8 +40,20 @@ def self.params_to_hash(query_params)
       params_hash
     end
 
-    def self.build_report_pms(query, table = "", targets = [])
+    def self.build_report_pms(query, table = "", targets = [], opts: {})
```
Drenmi, line 43: I think since the options are few and well known (i.e. not forwarded from some other interface) we could spell them out as keyword arguments.

```suggestion
    def self.build_report_pms(query, table = "", targets = [], result:, attach_csv:)
```

Also I think if we make the `result` object a bit richer, it could provide the `query` (and perhaps the `table` as well.)

WDYT?

spec/report_generator_spec.rb
```diff
@@ -130,5 +130,25 @@
       expect(result[1]["target_group_names"]).to eq([group.name])
       expect(result[2]["target_emails"]).to eq(["john@doe.com"])
     end
+
+    it "works with attached csv file" do
+      SiteSetting.personal_message_enabled_groups = group.id
+      DiscourseDataExplorer::ResultToMarkdown.expects(:convert).returns("le table")
```
Drenmi, line 136: Screenshot 2024-08-27 at 10 49 31 AM

Drenmi (approved):
LGTM! 🚀 

A few opportunities for improving the existing data model, but only suggestions, nothing that blocks this PR. 👍

app/controllers/discourse_data_explorer/query_controller.rb
```diff
@@ -179,49 +179,25 @@ def run
 
         render json: { success: false, errors: [err_msg] }, status: 422
       else
-        pg_result = result[:pg_result]
-        cols = pg_result.fields
+        filename =
```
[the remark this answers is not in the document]
Lhcfl: That's right. I renamed it to `content_disposition` 👍

lib/report_generator.rb
Lhcfl: Maybe? But I'm not sure. This result structure should be used in many places. Changing it will affect many files.

Lhcfl:
Thank you for your review! 🚀

PR #328, opened 2024-10-08
FEATURE: Add script to post report results in a topic regularly

martin-brennan:
This script is similar to the existing one that schedules
report results to to be sent to a PM on a regular basis,
but instead takes a topic ID and posts to that. This way
people can have report results sent to a public topic regularly
too and not have to deal with PM recipients and so on.

### DIFF
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -6,3 +6,6 @@ en:
       recurring_data_explorer_result_pm:
         title: Schedule a PM with Data Explorer results
         description: Get scheduled reports sent to your messages each month
+      recurring_data_explorer_result_topic:
+        title: Schedule a post in a topic with Data Explorer results
+        description: Get scheduled reports posted to a specific topic each month
--- a/config/locales/server.en.yml
+++ b/config/locales/server.en.yml
@@ -5,7 +5,34 @@ en:
     scriptables:
       recurring_data_explorer_result_pm:
         title: Schedule a PM with Data Explorer results
-        description: Get scheduled reports sent to your messages each month
+        description: Get scheduled reports sent to your messages
       recurring_data_explorer_result_topic:
         title: Schedule a post in a topic with Data Explorer results
-        description: Get scheduled reports posted to a specific topic each month
+        description: Get scheduled reports posted to a specific topic
+  data_explorer:
+    report_generator:
+      private_message:
+        title: "Scheduled report for %{query_name}"
+        body: |
+          Hi %{recipient_name}, your Data Explorer report is ready.
+
+          Query name:
+          %{query_name}
+
+          Here are the results:
+          %{table}
+
+          <a href='%{base_url}/admin/plugins/explorer?id=%{query_id}'>View query in Data Explorer</a>
+
+          Report created at %{created_at} (%{timezone})
+      post:
+        body: |
+          ### Scheduled report for %{query_name}
+
+          Here are the results:
+          %{table}
+
+          <a href='%{base_url}/admin/plugins/explorer?id=%{query_id}'>View query in Data Explorer</a>
+
+          Report created at %{created_at} (%{timezone})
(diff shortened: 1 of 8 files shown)


config/locales/server.en.yml
```diff
@@ -6,3 +6,6 @@ en:
       recurring_data_explorer_result_pm:
         title: Schedule a PM with Data Explorer results
         description: Get scheduled reports sent to your messages each month
+      recurring_data_explorer_result_topic:
+        title: Schedule a post in a topic with Data Explorer results
+        description: Get scheduled reports posted to a specific topic each month
```
SamSaffron, line 11: Each month? this feels weird ... regularly?

lib/discourse_data_explorer/report_generator.rb
```diff
@@ -73,6 +78,33 @@ def self.build_report_pms(query, table = "", targets = [], attach_csv: false, re
       pms
     end
 
+    def self.build_report_post(query, table = "", attach_csv: false, result: nil)
+      pms = []
+
+      upload = create_csv_upload(query, result) if attach_csv
+
+      post = {}
+      post["raw"] = "### Scheduled Report for #{query.name}\n\n" +
```
SamSaffron, line 87: this is a bit tricky, not localized at all, I guess it is just a caveat of the feature?

SamSaffron: Maybe we just remove almost all text except for a preamble - and translate that?

spec/automation/recurring_data_explorer_result_topic_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  context "when using recurring trigger" do
+    it "sends the post at recurring date_date" do
+      freeze_time 1.day.from_now do
+        expect { Jobs::DiscourseAutomation::Tracker.new.execute }.to change {
+          topic.reload.posts.count
+        }.by(1)
+
+        expect(topic.posts.last.raw).to include("Scheduled Report for #{query.name}")
```
SamSaffron, line 51: since we are generating stuff here ... might as well test that the query is rendered, even something basic like query is "select 'abctestabc'" and look for the text abctestabc?

additionally we need url support for our specific use case: 

eg: select `abc,https://test.com` as some_url 

data explorer has special handling for _url.

lib/discourse_data_explorer/report_generator.rb
martin-brennan: Yes you're right, I copy-pasted here and didn't follow my own rules about I18n :grimacing: Will have a look at the state of things, maybe we can translate the whole thing...

config/locales/server.en.yml
martin-brennan: Yeah you're right, will just remove "each month" from this and the other description, we already say schedule

spec/automation/recurring_data_explorer_result_topic_spec.rb
martin-brennan: Yep makes sense, will do that and handle URLs :+1:

martin-brennan: The URL special handling is on the client so we don't need to test for that here. For our internal example, the DE table generator returns something like this:

```
| team | in_progress_url | up_next_url | maybe_next_url | maybe_later_url | unprioritized_url |
| :----- | :----- | :----- | :----- | :----- | :----- |
| blah | 4,https://test.com/xyz | 13,https://test.com/xyz | 18,https://test.com/xyz | 15,https://test.com/xyz | 11,https://test.com/xyz |
```

So the client itself must wrap the count with the URL

martin-brennan:
@SamSaffron made the suggested changes and left a comment about the URL stuff :)

SamSaffron (approved):

SamSaffron:
thanks martin, let's give it a shot and see what Dave says.

PR #330, opened 2024-10-14
FIX: Optionally linkify URL columns server-side

martin-brennan:
Followup da1c99ec2d272a4eac8586a8b61794e8b755619b

We already had the functionality to convert results like
this:

```
|blah_url|
|--------|
|3,https://test.com|
```

To a link clientside, so in the UI it looks like this:

```
<a href="https://test.com">3</a>
```

With the addition of the recurring report to post automation,
and the existing report to PM automation, we also need to be
able to do this server-side, so reports don't come out malformed
if they have these type of count + URL columns.

Now if one of the automations generates the report, we tell the report
generator and markdown renderer to create an actual markdown link
for these URL columns.

### DIFF
--- a/lib/discourse_data_explorer/result_to_markdown.rb
+++ b/lib/discourse_data_explorer/result_to_markdown.rb
@@ -4,7 +4,7 @@ include HasSanitizableFields
 
 module ::DiscourseDataExplorer
   class ResultToMarkdown
-    def self.convert(pg_result)
+    def self.convert(pg_result, opts = {})
       relations, colrender = DataExplorer.add_extra_data(pg_result)
       result_data = []
 
@@ -17,7 +17,8 @@ module ::DiscourseDataExplorer
 
         row.each_with_index do |col, col_index|
           col_name = pg_result.fields[col_index]
-          related = relations.dig(colrender[col_index].to_sym) unless colrender[col_index].nil?
+          col_render = colrender[col_index]
+          related = relations.dig(col_render.to_sym) if col_render.present?
 
           if related.is_a?(ActiveModel::ArraySerializer)
             related_row = related.object.find_by(id: col)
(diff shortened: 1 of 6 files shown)


lib/discourse_data_explorer/result_to_markdown.rb
```diff
@@ -4,7 +4,7 @@
 
 module ::DiscourseDataExplorer
   class ResultToMarkdown
-    def self.convert(pg_result)
+    def self.convert(pg_result, opts = {})
```
Drenmi, line 7: Wondering, since we're passing only one option, and it's done explicitly, if we should use a keyword argument instead of an options hash here?

lib/discourse_data_explorer/result_to_markdown.rb
```diff
@@ -45,5 +49,16 @@ def self.convert(pg_result)
...
+
+    def self.guess_url(column_value)
+      split = column_value.split(/,(.+)/)
+
+      if split.length > 1
+        text = split[0]
+        url = split[1]
+      end
+
+      [url || column_value, text || column_value]
```
Drenmi, line 61: It's always a bit confusing to me to rely on variable hoisting (i.e. how come `text` and `url` aren't undefined). Technically we don't need the conditional. 🙂 

```ruby
text, url = column_value.split(/,(.+)/)

[url || column_value, text || column_value]
```

Drenmi (approved):
LGTM! 🚀

lib/discourse_data_explorer/result_to_markdown.rb
martin-brennan: Ah that is nicer, thanks!

lib/discourse_data_explorer/result_to_markdown.rb
martin-brennan: Oh yeah I meant to say, I wasn't really sure if `render_url_columns` was too specific or if we should have some generic `render_special_columns: true|false` option. But yes maybe for now we can just start with the keyword arg here

PR #346, opened 2024-12-09
UX: clarify the need for authorized extension

jjaffeux:
When checking `attach_csv` in the `Schedule a PM with Data Explorer results`
 automation script, `csv` has to be added to the list of authorized extensions in the site settings.

Screenshot 2024-12-09 at 17 09 14

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -82,7 +82,15 @@ after_initialize do
         field :query_id, component: :choices, required: true, extra: { content: queries }
         field :query_params, component: :"key-value", accepts_placeholders: true
         field :skip_empty, component: :boolean
-        field :attach_csv, component: :boolean
+        field :attach_csv,
+              component: :boolean,
+              validator: ->(attach_csv) do
+                if attach_csv && !SiteSetting.authorized_extensions.split("|").include?("csv")
+                  I18n.t(
+                    "discourse_automation.scriptables.recurring_data_explorer_result_pm.no_csv_allowed",
+                  )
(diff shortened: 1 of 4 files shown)


ZogStriP (commented):
Could you instead add a custom validation that will show the message if you enable this site setting but `CSV` isn't in the list of authorized extensions? 🤔

jjaffeux:
> Could you instead add a custom validation that will show the message if you enable this site setting but `CSV` isn't in the list of authorized extensions? 🤔

ah yes good idea, forgot I added this few months ago

plugin.rb
```diff
@@ -82,7 +82,15 @@ module ::DiscourseDataExplorer
         field :query_id, component: :choices, required: true, extra: { content: queries }
         field :query_params, component: :"key-value", accepts_placeholders: true
         field :skip_empty, component: :boolean
-        field :attach_csv, component: :boolean
+        field :attach_csv,
+              component: :boolean,
+              validator: ->(attach_csv) do
+                if attach_csv && !SiteSetting.authorized_extensions.split("|").include?("csv")
```
ZogStriP, line 88: I think we also need to handle the "*" which allows for "all" extensions.

jjaffeux: oh didn't know we do this...

jjaffeux: @ZogStriP ok I added the * case thanks

ZogStriP (approved):

PR #353, opened 2025-02-03
FIX: boolean field default state mismatch

renato:
When reading an existing state from the params defaults or from URL params, the input elements weren't representing the right state for booleans and "3-state/null booleans".

internal /t/-/118495

### DIFF
--- a/assets/javascripts/discourse/components/param-input-form.gjs
+++ b/assets/javascripts/discourse/components/param-input-form.gjs
@@ -194,10 +194,10 @@ export default class ParamInputForm extends Component {
       case "category_id":
         return digitalizeCategoryId(value);
       case "boolean":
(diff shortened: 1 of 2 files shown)


assets/javascripts/discourse/components/param-input-form.gjs
```diff
@@ -194,10 +194,10 @@ export default class ParamInputForm extends Component {
       case "category_id":
         return digitalizeCategoryId(value);
       case "boolean":
-        if (value == null) {
+        if (value == null || value === "#null") {
```
renato, line 197: the default `#null` wasn't working either

martin-brennan (approved):
Looks great, thanks for fixing this Renato!

PR #356, opened 2025-02-07
DEV: Split the Query Listing and Query Editing code

pento:
## ✨ What's This?

The code for listing all of the defined queries is mixed together with the code for editing a single query. Notably, this results in large amounts of unnecessary data being loaded for the list view, which causes substantial rendering slowdowns.

To address this issue, we now only load the necessary data for the list view, and load the full data when it's actually needed (any endpoint that returns a single query). The primary changes that achieve this are:

- Create a new `QueryDetailsSerializer` serialiser, which includes all of the query info, and change the existing `QuerySerializer` serialiser to only include the necessary attributes of each query for generating a list of them all.
- Split the monolith `/plugins/explorer` route into `/plugins/explorer` for showing just the list of queries, and `/plugins/explorer/queries/:query_id`, for showing/editing/running a specific query.

## 👑 Testing

Check that plugin functionality works as expected. I've gone through everything I could find, but I'm not 100% familiar with expected behaviour, so may've missed something.

Review the acceptance tests and system specs, to confirm that there are no missing uses cases.


app/serializers/discourse_data_explorer/query_details_serializer.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module ::DiscourseDataExplorer
+  class QueryDetailsSerializer < QuerySerializer
+    attributes :id,
```
martin-brennan, line 5: I think you can just add `:sql` here, all the other attributes should come from the base class, e.g.

https://github.com/discourse/discourse/blob/55316b9623cf772bec759e7392bc19c53e0de0e7/app/serializers/about_serializer.rb#L8-L10

assets/javascripts/discourse/templates/admin/plugins-explorer-index.hbs
```diff
@@ -0,0 +1,168 @@
+{{#if this.disallow}}
+  <h1>{{i18n "explorer.admins_only"}}</h1>
+{{else}}
+  {{#unless this.validQueryPresent}}
```
martin-brennan, line 4: I don't think you need this check anymore, it's been removed from the JS and it was only relevant when selecting a query on the main page

config/locales/server.en.yml
```diff
@@ -23,7 +23,7 @@ en:
           Here are the results:
           %{table}
 
-          <a href='%{base_url}/admin/plugins/explorer?id=%{query_id}'>View query in Data Explorer</a>
+          <a href='%{base_url}/admin/plugins/explorer/queries/%{query_id}'>View query in Data Explorer</a>
```
martin-brennan, line 26: Maybe just check on Meta if we have Data Explorer docs there after you merge this, and if so update the links.

It's also likely that people have bookmarked these in a lot of places, so I think it's probably worth doing a 302 redirect on the server-side to go to the new endpoint

martin-brennan: Or if there was no server-side thing, perhaps do a clientside only redirect if they land on the /explorer route with an id?

martin-brennan (approved):
Looks great, mostly a vibes based review and local testing, I think we should be able to fix any issues that crop up fairly quick, and any potential bugs are outweighed by the benefit of doing this. Just a few minor comments.

config/locales/server.en.yml
pento: Yep, it was only client side, so I've added a redirect there.

pento: I've also had a read through the Data Explorer docs and tutorials on Meta, I couldn't see any references to the `?id=123` URL structure.

martin-brennan:
Thank you!

PR #359, opened 2025-02-13
FIX: Query downloads were being passed an incorrect query object.

pento:
## ✨ What's This?

This is a follow-up to #356.

The previous PR missed changing the name of the query object when passing it to `QueryResultsWrapper`, which resulted in the download links not working properly after a query was run.

This change fixes that bug, and includes an acceptance test to ensure it stays fixed during future work on this plugin.

## 👑 Testing

- Visit `/admin/plugins/explorer`
- Click on any query.
- Click the **Run** button.
- Click the **CSV** and **JSON** buttons, make sure the download happens as expected.

### DIFF
--- a/test/javascripts/acceptance/run-query-test.js
+++ b/test/javascripts/acceptance/run-query-test.js
@@ -1,11 +1,20 @@
 import { click, visit } from "@ember/test-helpers";
 import { test } from "qunit";
+import sinon from "sinon";
 import { acceptance } from "discourse/tests/helpers/qunit-helpers";
 import { i18n } from "discourse-i18n";
 
 acceptance("Data Explorer Plugin | Run Query", function (needs) {
   needs.user();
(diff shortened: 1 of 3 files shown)


test/javascripts/acceptance/run-query-test.js
```diff
@@ -233,6 +248,74 @@ acceptance("Data Explorer Plugin | Run Query", function (needs) {
     assert.dom("canvas").exists("the chart was rendered");
   });
 
+  test("runs query and is able to download the results", async function (assert) {
+    await visit("/admin/plugins/explorer/queries/-6");
+
+    await click("form.query-run button");
+
+    const createElement = document.createElement.bind(document);
```
martin-brennan, line 256: Personally I would have written this as a system spec to avoid needing to do all these mocks  and stubs, but it's cool you got it to work this way 👍

martin-brennan (approved):

PR #363, opened 2025-03-13
FIX: Discourse automation reports with parameters

Drenmi:
### What is the problem?

When setting up an automation to create a DM or a post with a report on a recurring basis, and using a Data Explorer query that has parameters, we encounter an error.

### Why is this happening?

The `.params_to_hash` currently expects an array of arrays for the parameters, e.g.:

```ruby
[["group_name", "admins"]]
```

but in reality they seem to be arrays of hashes:

```ruby
[{ "key" => "group_name", "value" => "admins" }]
```

### How come we only found out now?

We do have tests for the report generator, and we do pass query parameters to it, but 1) we hard-code the parameters as an array of arrays and 2) the parameters aren't in the query used as a fixture.

### How does this fix it?

This change makes `ReportGenerator.params_to_hash` work with arrays of hashes. It also preserves the ability to work with nested arrays in case this is used somewhere else.

### DIFF
--- a/lib/discourse_data_explorer/report_generator.rb
+++ b/lib/discourse_data_explorer/report_generator.rb
@@ -35,26 +35,10 @@ module ::DiscourseDataExplorer
       build_report_post(query, table, attach_csv: opts[:attach_csv], result:)
     end
 
-    private
-
     def self.params_to_hash(query_params)
       params = JSON.parse(query_params)
-      params_hash = {}
 
-      if !params.blank?
-        param_key, param_value = [], []
-        params.flatten.each.with_index do |data, i|
-          if i % 2 == 0
-            param_key << data
-          else
-            param_value << data
-          end
-        end
-
-        params_hash = Hash[param_key.zip(param_value)]
-      end
-
-      params_hash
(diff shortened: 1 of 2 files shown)


lib/discourse_data_explorer/report_generator.rb
```diff
@@ -35,26 +35,10 @@ def self.generate_post(query_id, query_params, opts = {})
       build_report_post(query, table, attach_csv: opts[:attach_csv], result:)
     end
 
-    private
```
Drenmi, line 38: Scoping keywords put in the class body don't affect class methods. All methods in this file are already public.

lib/discourse_data_explorer/report_generator.rb
```diff
@@ -35,26 +35,10 @@ def self.generate_post(query_id, query_params, opts = {})
...
-        params.flatten.each.with_index do |data, i|
-          if i % 2 == 0
-            param_key << data
-          else
-            param_value << data
-          end
-        end
-
-        params_hash = Hash[param_key.zip(param_value)]
-      end
```
Drenmi, line 55: This chunk is essentially just `Array#to_h`.

jjaffeux (approved):

PR #372, opened 2025-04-23
DEV: improve group query access for report PM

dbattersby:
This change updates the report generator script to handle groups first,  that way when the option `users_from_group: true` is used we can determine query access based on the user rather than the whole group.

**Example:**
The data explorer query group is set to all registered users, but the automation script has a smaller subgroup of users. Those users should still have access as they are part of both the smaller group and the larger (all users) group, despite the subgroup not having explicit access.

### DIFF
--- a/lib/discourse_data_explorer/report_generator.rb
+++ b/lib/discourse_data_explorer/report_generator.rb
@@ -127,11 +127,26 @@ module ::DiscourseDataExplorer
 
       query_group_ids = [Group::AUTO_GROUPS[:admins]].concat(query.groups.pluck(:group_id)).uniq
 
-      groups.each do |group|
-        if users_from_group
-          users = (users + group.users).uniq
-        elsif query_group_ids.include?(group.id)
-          result << [group.name, "group_name"]
+      if users_from_group
+        result.concat(
+          User
+            .joins(:group_users)
+            .where(group_users: { group_id: groups.ids })
+            .where(
+              "users.admin OR EXISTS (
+                SELECT 1 FROM group_users gu
+                WHERE gu.user_id = users.id
+                AND gu.group_id IN (?)
+              )",
+              query_group_ids,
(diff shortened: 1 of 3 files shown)


lib/discourse_data_explorer/report_generator.rb
```diff
@@ -125,20 +125,20 @@ def self.filter_recipients_by_query_access(recipients, query, users_from_group:
...
 
       groups.each do |group|
-        if group.id == Group::AUTO_GROUPS[:admins] || query.query_groups.exists?(group_id: group.id)
-          if users_from_group
-            result.concat(group.users.pluck(:username).map { |username| [username, "username"] })
-          else
-            result << [group.name, "group_name"]
-          end
+        if users_from_group
+          users = (users + group.users).uniq
```
tgxworld, line 132: This feels dangerous to me performance wise because we are going to be loading every single user record from the group into memory. Do you know if we have any limits upstream of this method?

dbattersby: Good point. Maybe we can batch this when validating, something like:

```
group.users.find_each(batch_size: 1000) do |group_user|
  if Guardian.new(group_user).user_can_access_query?(query)
    result << [group_user.username, "username"]
  end
end
```

tgxworld: Even with batching, we are still vulnerable if the group has alot of users that can access this query. That being said, this was bad even before your change since we plucked all the usernames from the group. I wonder if we can achieve the following method in a single query instead of having to load ActiveRecord objects and iterate through it.

```
  add_to_class(:guardian, :user_can_access_query?) do |query|
    return false if !current_user
    return true if current_user.admin?
    query.groups.blank? || query.groups.any? { |group| user_is_a_member_of_group?(group) }
  end
 ```

